PR #31, opened 2019-09-26
Replace inner make_prior loop with np version

hyanwong:
I assume this should create identical results. Since it's the innermost loop, I'm hoping it will speed things up a bit (e.g. the initial make_prior), although this is another possibility for caching too.

Do the unit tests now include a simple testcase to check inference accuracy doesn't change?

### DIFF
--- a/tsdate/date.py
+++ b/tsdate/date.py
@@ -88,9 +88,7 @@ class prior_maker():
         Gives expectation of tau squared conditional on m
         Equation (10) from Wiuf and Donnelly (1999).
         """
-        t_sum = 0
-        for t in range(m, n + 1):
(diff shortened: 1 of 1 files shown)


awohns:
Some simple tests of inference accuracy are in 'TestPrebuilt' in 'tests/test_inference.py'. I haven't modified them to test the accuracy of the approximate prior though, I should do this

hyanwong:
Right, I see them.

I'm guessing that you want to merge this?

It would be easy to numpy-ize the` i==n` branch of `tau_var()` too, which would be useful since it is even used in `tau_var_lookup()`

hyanwong:
Can't see why I shouldn't merge this

PR #32, opened 2019-09-26
Create bespoke precalculated priors & test

hyanwong:
@awohns - this is quite a bit of extra hacking on your precalculated prior code. Please check it out

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -23,6 +23,7 @@
 Test cases for the python API for tsdate.
 """
 import unittest
+import os
 
 import tskit  # NOQA
 import msprime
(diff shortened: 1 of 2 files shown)


hyanwong:
Explanation: for testing, we don't want to make the precalculated approx prior with 1000 tips, as this takes too long. So we have a variable `precalc_approximation_n` set to 1000. When we precalculate the prior, we use this variable to set the value for the stored file. So in tests, we can change it after instantiation, then fire off the `precalculate_prior_for_approximation()` function, which will make a smaller example.

I've also embedded the number in the filename, so that we don't end up overwriting the real one.

hyanwong:
I've stuck some template code in test_functions() for looking to check that the approximation is being used when n=10. It's not checking anything yet, and I might have code the logic wrong, but at the moment it's giving 

```
tsdate/date.py:44: RuntimeWarning: divide by zero encountered in double_scalars
```

Which is worth investigating.

awohns:
I’ve just had a look at this, can’t tell where that runtime warning is coming from when I run the test one by line, but it’s obviously from somewhere in 'test_precalculated_prior'

awohns:
Also, is it worth trying to test the accuracy of the approximation with unit tests?

hyanwong:
If it takes 5 mins to calculate the approximate prior for n=1000, then that's not a unit-test thing (you could make it a "functional test", and roll it into the evaluation testing framework instead of the unit tests). You want unit tests to take fractions of seconds.

But it's worth testing that the logic behind the interpolation etc is working - i.e. that when you ask for an approximate value, you pick out the correct value from the precalculated table. For example, at the moment you do `self.prior_df.iloc[self.prior_df.index.searchsorted(i / n)].values[0]` which I think simply picks the nearest value: I wonder if you should interpolate instead, which should be almost as quick (take the weighted mean of the 2 values on either side). This would need some testing with real values: around the code on line 226 of test_functions.py where I've left a comments but haven't actually coded anything up.

hyanwong:
> I’ve just had a look at this, can’t tell where that runtime warning is coming from when I run the test one by line, but it’s obviously from somewhere in 'test_precalculated_prior'

It probably only gets exposed when doing the approximation with small n (e.g. `precalc_approximation_n` = 10). It's a bit horrible to have the functionality depend on n like this, so worth checking out (you could debug by hardcoding n=10 rather than n=1000 into your existing code and seeing how the approximation does)

PR #34, opened 2019-09-26
Make tau_var faster for i==n

hyanwong:
As mentioned in https://github.com/awohns/tsdate/pull/31#issuecomment-535503171


hyanwong:
Can't see why I shouldn't merge this

PR #47, opened 2019-11-28
Put all the prior dataframes into a single obj

hyanwong:
And make find_node_tips into a class.

There is some substantial renaming going on here. I hope that's OK @awohns . Do suggest better names for functions and classes if you can think of any.


awohns:
Lots of awesome-looking changes in here!! I'll review it tomorrow. Might be easiest to go over it together if you'll be in?

hyanwong:
Yep, I'm in tomorrow. Chat then.

PR #57, opened 2019-12-02
Node tip=plus forward backward optim

hyanwong:
For testing only

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -373,7 +373,8 @@ class TestPriorVals(unittest.TestCase):
         priors.add(ts.num_samples, approximate=False)
         grid = np.linspace(0, 3, 3)
         mixture_prior = tsdate.get_mixture_prior(span_data, priors)
-        prior_vals = tsdate.get_prior_values(mixture_prior, grid, ts)
+        nodes_to_date = span_data.nodes_to_date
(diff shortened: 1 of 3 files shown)


awohns:
Investigating the issue with the node tip weights:
Run the following:
`ts = msprime.simulate(sample_size=10, length=2e6, Ne=10000, mutation_rate=1e-8,
                          recombination_rate=1e-8, random_seed=11)`
Look at node 703. It has a value of nan in mn_post, forwards, backwards etc.
We see that the issue is that it is listed as having zero tips:
`span_data.weights(703)`
`{10: Weights(descendant_tips=array([10,  0]), weight=array([0.48817349, 0.51182651]))}`
The actual value should be this (the old code returns the right value):
`{10: Weights(descendant_tips=array([10]), weight=array([1.]))}`

hyanwong:
> Investigating the issue with the node tip weights: ...
> Look at node 703. It has a value of nan in mn_post, forwards, backwards etc.
> Let me know if you can reproduce this @hyanwong

Hi - I've now fixed this in the latest commit in this branch @awohns, and added the example to the test cases (it was when root in the first tree dropped out in the second - I was forgetting to initialise the number of children for all parent nodes at the start of the iteration loop). Let me know if you find any other bugs.

hyanwong:
Tested a little, closing now that https://github.com/awohns/tsdate/pull/54 is merged

PR #58, opened 2019-12-02
Save a backup of the approx cache for testing

hyanwong:


hyanwong:
And also .gitignore the standard created file.

PR #59, opened 2019-12-02
Parallel poisson pmf

hyanwong:
For reference as an implementation of multiprocessing

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -373,7 +373,8 @@ class TestPriorVals(unittest.TestCase):
         priors.add(ts.num_samples, approximate=False)
         grid = np.linspace(0, 3, 3)
         mixture_prior = tsdate.get_mixture_prior(span_data, priors)
-        prior_vals = tsdate.get_prior_values(mixture_prior, grid, ts)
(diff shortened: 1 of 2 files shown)


hyanwong:
This should be perf tested, as (because the `get_ll` function is reasonably fast, but gets called millions of times) it may be that the overhead of pickling the function outweighs the benefit from parallelisation. I don't understand well enough the way multiprocessing works (especially with functools.partial) to know if we will see a substantial increase in speed or not, but I'm hopeful...

hyanwong:
Premature optimisation and all that, but I suspect that my recent implementation using a set (`result ={...}`) will be just as slow, as there will be hash collisions when making the set, rather than making the dictionary. 

Perhaps better would be to put the number of mutations and span into a basic numpy [structured array](https://docs.scipy.org/doc/numpy/user/basics.rec.html) and use np.unique to find the keys for the `result` array.

awohns:
It's unclear to me if this is actually speeding up the pmf calculation... preliminary testing shows that using num_threads=5 is often slower than the single threaded version

hyanwong:
Done as part of https://github.com/awohns/tsdate/pull/77

Issue #65, opened 2019-12-04
Add support in backwards step for dangling nodes


hyanwong:
We can put priors on *some* dangling nodes, and therefore date them because they may not be dangling in some trees. But there may be a few nodes that have no descendant tips in any tree. 

We can, however, date these by not having a prior at all, and simply using the likelihood from the number of mutations and edge span. This is something to do on the backwards step, and I’ll adjust get_weights(u) to return None (or maybe an empty dict) so it can be intercepted and dealt with.

hyanwong:
Near dupe of #38

Issue #38, opened 2019-10-02
Dating dangling nodes 


awohns:
We currently can't date an unsimplified tree sequence with dangling nodes. We might want to think about how to do this, the problem is that there isn't a clear prior for these nodes.

hyanwong:
To be clear, we might want to leave dangling edges in place because they are useful attachment points for augmented samples. Unfortunately they may be hard (impossible?) to date because of the lack of prior

hyanwong:
It should be possible to assign the global prior to these nodes now

hyanwong:
Not longer an issue for `variational_gamma`.

PR #67, opened 2019-12-05
Start addressing #65

hyanwong:
We still need to deal with NaNs in the forwards matrix when calling `backward_algorithm()`. These should mark dangling nodes, although I'm slightly worried that other bugs might cause NaNs too, so perhaps that isn't such a good convention to establish. Other suggestions welcome (perhaps np.inf?)

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -86,11 +86,17 @@ class TestNodeTipWeights(unittest.TestCase):
                 if not tree.is_sample(n):
                     nonsample_nodes.add(n)
         self.assertEqual(set(span_data.nodes_to_date), nonsample_nodes)
+        dangling_only_nodes = set(span_data.dangling_only_nodes())
         for focal_node in span_data.nodes_to_date:
+            weight = 0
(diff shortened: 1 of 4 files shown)


hyanwong:
Just revisiting this. We now have a specific tskit Nan: `tskit.UNKNOWN_TIME`, so we should be using this for dangling nodes, if we are still thinking of doing this (I'm not sure if it's relevant any more)

awohns:
Right now we error out if we see dangling nodes. #38 and #65 are marked on the "wish list". Definitely seems like something that we _should_ do. Perhaps in release 1.1

hyanwong:
> Right now we error out if we see dangling nodes. #38 and #65 are marked on the "wish list". Definitely seems like something that we _should_ do. Perhaps in release 1.1

Yes, perfect. It actually helps us a lot that there is a *specific* NaN we can use to mean "unknown", which we won't confuse with NaNs produced for other reasons.

hyanwong:
Closing this due to inactivity: we'll end up implementing something better for historical samples using variational_gamma anyway.

PR #72, opened 2019-12-06
Remove separate norm constants

hyanwong:
I'm not being stupid here am I @awohns ? It looks like you don't need to save the normalising constants, but can just normalise them on assignment like the attached. Tests all pass...


awohns:
Yes can definitely be removed, I don’t think we’d ever need them for anything. forgot that was still in there!

PR #77, opened 2019-12-09
Create a class to abstract the lower triangular L matrix

hyanwong:
This seems to pass the test but still complains about "date.py:1079: RuntimeWarning: divide by zero encountered in log  log_backwards = np.log(backwards)" - it might do some more than it used to, as I removed some of the (probably needless) stuff in the normalisation of the backwards (i.e., I simply omitted backwards[X, 0] from all normalisation. I guess this is OK, but maybe you could check that I haven't messed up, @awohns ? And also that the changes make sense, and the new methods are named sensibly (I suspect some like `expand_forwards` could do with better names)

### DIFF
--- a/tsdate/date.py
+++ b/tsdate/date.py
@@ -26,6 +26,8 @@ from collections import defaultdict, namedtuple
 import logging
 import os
 import itertools
+import multiprocessing
+import functools
 
 import tskit
 
@@ -37,6 +39,10 @@ from tqdm import tqdm
 
 FORMAT_NAME = "tsdate"
 FORMAT_VERSION = [1, 0]
+
+# Hack: monkey patches to allow tsdate to work with non-dev versions of tskit
+# TODO - remove when tskit 0.2.4 is released
+tskit.Edge.span = property(lambda edge: (edge.right - edge.left))  # NOQA
 
 
 def gamma_approx(mean, variance):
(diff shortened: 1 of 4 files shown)


tsdate/date.py
```diff
@@ -993,41 +1051,33 @@ def backward_algorithm(
...
-            norm[edge.child] = max(backwards[edge.child, :])
-            backwards[edge.child, :] = \
-                np.divide(backwards[edge.child, :], norm[edge.child])
-            backwards[edge.child, :][np.isnan(backwards[edge.child, :])] = 0
-    backwards_log = np.log(backwards)
-    posterior = np.exp(forwards + backwards_log)
+                    vv = lls.sum_tri_cols(dt)
+            backwards[edge.child, 1:] = vv[1:] / max(vv[1:])
+            backwards[edge.child, np.isnan(backwards[edge.child, :])] = 0
+    log_backwards = np.log(backwards)
```
awohns: You can add  
`with np.errstate(divide='ignore'):`
above this line and it won't complain.

awohns:
Otherwise I think this looks great!

tsdate/date.py
hyanwong: Well, it's not so much that I am worried about h error, it's that I'm worried that I'm triggering it more often than before (I didn't have much of a feel for how much that was, though). I think it would be better to solve the error rather than hide it, though. You have a comment above that implies as much. What's the problem here?

awohns:
The log is only called once at the end of the forward and once at the end of the backward algorithms. So I don’t think you could be calling it more than before as there will always be zeros in the forward and backward matrices. I don’t think this is an issue, it’s a warning rather than an error after all… I think we just have to handle the resulting -Inf values correctly.

hyanwong:
Can I merge this @awohns? There's quite a lot of renaming etc (up/down not forwards/back, lower vs upper tri, etc etc.) so I thought I should check with you before merging?

awohns:
Just had a look and this seems good to me

hyanwong:
Closes https://github.com/awohns/tsdate/issues/75
Probably closes https://github.com/awohns/tsdate/issues/73
Closes https://github.com/awohns/tsdate/issues/71 (although not this bit about storing only the undated nodes)

hyanwong:
> Just had a look and this seems good to me

OK, merging. You might want to rebase your ongoing work onto master, as lots will have changed

PR #80, opened 2019-12-10
added support for internal samples, more tests coming

awohns:
Supports internal nodes. Creates a new class to store the identity and time of ancient and modern nodes. Upward and backward algorithms properly handle these fixed nodes (skipping them in the recursion). One test so far of how internal nodes are handled in the upward algorithm, will add more now.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -528,6 +528,7 @@ class TestUpwardAlgorithm(unittest.TestCase):
         grid = np.array([0, 1.2, 2])
         mixture_prior = tsdate.get_mixture_prior(span_data, priors)
         nodes_to_date = span_data.nodes_to_date
+        fixed_nodes = tsdate.FixedSamples(ts, grid)
         prior_vals = tsdate.get_prior_values(mixture_prior, grid, ts, nodes_to_date)
         theta = 1
         rho = None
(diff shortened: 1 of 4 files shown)


awohns:
I've had a go at changing the SpansBySamples to handle internal nodes, but I've ended up breaking @hyanwong. The change is conceptually pretty simple, we just have to keep track of modern samples rather than all samples. However, in save_to_span_data() and set_spans_connected_unary(), set_spans_unconnected_unary(), and finalise(), you use the `tree.num_samples()` function. This counts samples which are _internal_, in addition to tips. Is there an easy switch we can make here? I can't easily find something in tskit that does what we want.

awohns:
Just to emphasise, this PR is _not_ ready to merge, number of changes still necessary, primarily handling the spans correctly.

hyanwong:
> Is there an easy switch we can make here?

I think there's a constant time "tracked samples" counter isn't there? So we should be able to set the tracked samples as all the contemporaneous ones and use `num_tracked_samples()`. We have to be a little careful here with missing data, as isolated samples are probably reported too, though.

hyanwong:
I think this is doing it slightly wrong:

1) Once https://github.com/awohns/tsdate/pull/81 is merged we shouldn't have to set ancient samples to the nearest time in the grid. Instead we can calculate the time deltas exactly.
2) We should be able to perform the upwards pass from ancient samples too: it's just that the nodes above them may not have a prior. What we should think about is how to set the prior on these nodes, if at all possible. If not, I guess we could, for the moment, use a uniform prior instead.

PR #81, opened 2019-12-11
Only store hidden states for unknown nodes

hyanwong:
And don't bother calculating a full lower tri matrix for fixed nodes.

Also put upward/downward in a class of their own, to keep a handle on node mappings etc.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -442,7 +442,6 @@ class TestLikelihoodClass(unittest.TestCase):
     def test_get_mut_edges(self):
         ts = utility_functions.two_tree_mutation_ts()
         mutations_per_edge = tsdate.Likelihoods.get_mut_edges(ts)
-        print(mutations_per_edge)
         for e in ts.edges():
             if e.child == 3 and e.parent == 4:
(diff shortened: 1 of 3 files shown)


hyanwong:
There are some bugs in here @awohns, so that tests aren't passing. But I thought I would make the first PR so you could see what I'm suggesting doing.

One advantage is that we don't need to calculate the whole lower triangular matrix for edges from a fixed node - we only need the delta-times from the fixed position. This should save some time, although perhaps most edges aren't sample edges.

A better reason for doing this is that we can then have a more sensible likelihood calc for non-contemporaneous nodes.

hyanwong:
This now all works fine. How do we want to proceed @awohns ? I guess I could talk it over with you tomorrow (Thu), unless you are planning to work on any more tsdate stuff tonight, in which case the sooner we merge it the better (less work reconciling divergent branches)

awohns:
Hi Yan,

Perhaps we can talk this over tomorrow, maybe in person if that’s easiest (or by GitHub or phone, I’m flexible). I’m taking my drivers theory test at 11am so I’ll be in town.

PR #83, opened 2019-12-12
Only tally fixed nodes for prior if at time 0

hyanwong:
This seems pretty clear, and obvious to merge. No tests for non-contemporaneous samples yet, though...


hyanwong:
See https://github.com/awohns/tsdate/pull/80#issuecomment-564179257

PR #90, opened 2020-01-10
Merging geometric with yan

hyanwong:
This now passes all but 2 tests. One test fails because of an underflow/overflow error (`tests/test_inference.py:TestSimulated.test_simple_sim_larger_example`), the other because a hard coded test value probably needs updating. Otherwise, I think it can probably be merged into master, so that at least we have a geometric scaling algo there. What do you think, @awohns ?

### DIFF
--- a/evaluation/evaluation.py
+++ b/evaluation/evaluation.py
@@ -170,7 +170,7 @@ def run_vanilla_simulation(
     return ts
 
 
-def out_of_africa():
+def out_of_africa(sample_size, mutation_rate, recombination_rate, length):
     # First we set out the maximum likelihood values of the various parameters
     # given in Table 1.
     N_A = 7300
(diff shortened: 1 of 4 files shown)


awohns:
This version overflows because it isn’t solely in log-space. I need to merge it with ‘full-log-geo-scale’ which I’ll do tomorrow. Sorry!

hyanwong:
Yes, I guessed that was the problem. Happy to wait for your update. I hope it's not too difficult to merge all these together. It would be useful for us all to be working from the same Master branch, I think.

hyanwong:
Did you manage to merge with the log-geoscale branch, @awohns ?

awohns:
I’m working on it, there are a few tests that don’t pass because everyone is kept “logged” all the time and require some fiddly and tedious reworking unfortunately...

hyanwong:
No hurry. I was just wondering.

PR #97, opened 2020-02-19
Fixes and Polishing for Beta Release

awohns:
Remaining points:

- [x] Write tutorial and host docs
- [x] Add some evaluation plots


awohns:
I also need to check the input more thoroughly (make sure the rec and mut rates are floats etc.)

awohns:
Running on 1000g chromosome 20 is now _extremely_ fast (30 mins, excluding the constrain ages step). However, I can't get the numerical stability worked out for the inside outside, there's overflow on the outside regardless of the normalisation strategies I've tried. The maximization works fine though. I think I'll change the default to maximization given this issue. Looking at the evaluation figures, the maximization often isn't that much worse than inside-outside anyway

PR #106, opened 2020-02-26
Bomb out if dangling nodes, and warn if unary nodes present

hyanwong:
As discussed with @awohns


awohns:
This looks perfect, I say go ahead and merge!

hyanwong:
I'll add another test just to get the coverage up , then merge.

PR #107, opened 2020-02-26
Correct outside traversal

awohns:
Change the node ordering in the outside pass and outside maximization algorithms. Add an example tree where this is an issue as well as a simulation/inference check.


awohns:
What do you think about this @hyanwong ?

hyanwong:
Looks good to me, but you can probably squash all the commits into one before merging

PR #108, opened 2020-02-26
Add warning about creating the user cache

hyanwong:
Still needs @awohns to convert to use appdirs.user_cache_dir


hyanwong:
Obsoleted by #112

PR #109, opened 2020-02-26
Caching issue

awohns:
Not passing because it looks like we might need to create the cache dir if it doesn't already exist?

### DIFF
--- a/tsdate/date.py
+++ b/tsdate/date.py
@@ -29,6 +29,8 @@ import itertools
 import multiprocessing
 import operator
 import functools
+import pathlib
+import appdirs
 
 import tskit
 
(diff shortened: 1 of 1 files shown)


hyanwong:
> Not passing because it looks like we might need to create the cache dir if it doesn't already exist?

Yes, stdpopsim does it using os.makedirs: https://github.com/popsim-consortium/stdpopsim/blob/28fb59f2cbd16794de257c5fda15d143c547853e/stdpopsim/genetic_maps.py#L102

awohns:
We now create a cache directory, passes tests too. What do you think @hyanwong ?

hyanwong:
Need to go now, but merge if you think it's OK.

PR #111, opened 2020-02-26
Properly detect dangling / unary nodes

hyanwong:
This corrects my previous mistake (simplify() can re-order the node numbers, even if everything else is unchanged)


hyanwong:
Just realised that this isn't enough, actually. It doesn't catch the case when a node is partially dangling.

Probably worth merging, though, and I can work on better logic later.

PR #121, opened 2020-02-28
Use the theta & rho from the likelihood class

hyanwong:


hyanwong:
As discussed. Better coverage now, so merging

PR #122, opened 2020-02-28
Use ts from likelihood class, and allow dangling nodes in prior only

hyanwong:
This only issues a warning (rather than raising an error) if we attempt to calculate a prior with dangling nodes. The prior is then set to NaN for a dangling node, which should be fine.

It's now only when we try to run the Inside algo that we fail on a dangling node. I think this is the correct way to do it.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -1217,14 +1217,6 @@ class TestBuildPriorGrid(unittest.TestCase):
         self.assertRaises(ValueError, tsdate.build_prior_grid, ts,
                           prior_distribution="foobar")
 
-    def test_warn_dangling(self):
-        ts = utility_functions.single_tree_ts_n3_dangling()
-        with self.assertLogs(level="WARNING") as log:
-            tsdate.build_prior_grid(ts)
-            self.assertEqual(len(log.output), 2)
-            self.assertIn("dangling", log.output[0])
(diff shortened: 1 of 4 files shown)


hyanwong:
This change required a fair bit of reworking, @awohns. Should I merge, or do you want to review?

awohns:
Awesome!
If you don’t need it merged right away, I’d like to have a look beforehand. I can review it tomorrow if that works

hyanwong:
>  If you don’t need it merged right away, I’d like to have a look beforehand. I can review it tomorrow if that works

Yep, no hurry on this. Review whenever you like. I'll leave it unmerged

awohns:
This also looks good to me, I'll have a look at the merge conflicts now

hyanwong:
I'm happy to have a go at resolving, then you can review

hyanwong:
This is ready for review, @awohns . There's a bit of a hack in `fill_prior` where we construct a list of `datable_nodes` by picking everything that isn't a sample. We need to do this because we are allowing dangling nodes to be represented in the prior using a row of NaNs, and we want to temporarily "pretend" these are datable (and allocate them hidden state likelihoods), then crap out in the inside pass.

The idea here is that we could potentially assign probabilities for hidden states on a dangling node e.g. on the downwards pass, but simply haven't implemented it yet.

PR #123, opened 2020-02-29
Switch to only storing fixed node data in Prior

hyanwong:
I realised belatedly that fixed nodes do not need to store likelihood values for the prior, but *times* - i.e. all the likelihood is concentrated at a precise time (not just a time slice, but a floating point time value). Additionally, we don't want to change this during the inside or outside pass. So I've changed the NodeGridValues class to take out the fixed node stuff, and created a new `Prior` subclass that adds the ability to store a time (*not* a likelihood) for fixed nodes.

This is not for merging yet - it is the first iteration of incorporating non-contemporaneous samples into tsdate.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -33,10 +33,10 @@ import msprime
 import tsinfer
 
 import tsdate
-from tsdate.date import (SpansBySamples, PriorParams,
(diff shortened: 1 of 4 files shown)


hyanwong:
Closing, as we aren't really focussed on making priors with v_gamma.

PR #125, opened 2020-03-03
Remove noncontemporaneous samples from prior

### LINKED ISSUE #124: Remove non-contemporaneous samples before calculating prior
I've just realised that some nodes which are coalescent nodes in a normal tree will turn out to be unary when calculating the conditional coalescent prior. E.g. 

```
    5
   / \
  |   4
  |   |\
  |   |[2]
  |   |
 [0] [1]
```

Here node 4 is a coalescent node, but if we are calculating the prior only using samples at time 0, it turns into a unary node.

I think the simplest fix here is to `simplfy(contemporary_node_list... keep_unary=True)` for the TS that we use when calculating the prior, and then map nodes in the simplified TS against ones in the TS that we finally use for the InOut algorithm. What do you think @awohns?

hyanwong:
Fixes #124 - I think we should simply merge this, since all tests pass. Then I can get on with replumbing the NodeGridPrior class, which will allow us to deal with noncontemporaneous samples properly.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -425,9 +425,7 @@ class TestPriorVals(unittest.TestCase):
         priors.add(ts.num_samples, approximate=False)
         grid = np.linspace(0, 3, 3)
         mixture_prior = priors.get_mixture_prior_params(span_data)
-        nodes_to_date = span_data.nodes_to_date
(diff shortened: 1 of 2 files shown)


awohns:
LGTM!

PR #130, opened 2020-05-06
Support for keep_intervals and delete_intervals

awohns:
One line change to support the tree sequences that result from a call to keep_intervals() or delete_intervals().

Tests to show that the priors are unchanged.

NOTE: the resulting date estimates will be different with _any_ deletion of an internal interval. The test "test_intervals()" in tests/test_inference.py() demonstrates this. If we take an internal interval in the two_tree_ts_extra_long() of [0.5, 1.0], the resulting dates will be different than the results of tsdate on two_tree_ts(). I think this is expected.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -195,6 +195,43 @@ class TestNodeTipWeights(unittest.TestCase):
             self.assertGreater(len(log.output), 0)
             self.assertIn("dangling", log.output[0])
 
+    def test_single_tree_n2_delete_intervals(self):
+        ts = utility_functions.single_tree_ts_n2()
+        deleted_interval_ts = ts.delete_intervals([[0.5, 0.6]])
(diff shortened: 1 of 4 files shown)


hyanwong:
I haven't looked yet, but shouldn't the priors change if there used to be an edge (& sample nodes) in the deleted section?

I think one test should be that a set of dates obtained from a TS with delete_interval called on a fraction at the start and end should be exactly the same as when trim() has then been called on the delete_interval ts.

awohns:
Yep, good point, the priors should change if there is more than one tree and there used to be an edge in the deleted section. If there's only one tree, the priors should remain unchanged.
For the >1 tree case, the tests check that removing or keeping intervals in a larger tree sequence  creates the same priors and end result as inputting the smaller tree sequence to begin with.

Basically, the only change to the priors code is not to add a priors to account for trees where there are zero samples.

Issue #100, opened 2020-02-24
constrain_ages_topo is very slow


awohns:
constrain_ages_topo() takes over an hour to run on the 1000G chromosome 20 tree sequence, while the rest of the inference only takes 30 minutes.

awohns:
Fixed in #132

PR #131, opened 2020-07-31
Faster constrain ages with tests

awohns:
As mentioned in #100, constrain_ages_topo is very slow. This PR has a new implementation with is much faster. A number of tests are included.


awohns:
Will make another PR without the black formatting changes, which confuses everything.

PR #136, opened 2020-10-26
Support for ancient samples

awohns:
Adds two functions to support the inference of tree sequences from modern and ancient samples. The first is `sites_time`: a function to retrieve the times associated with each variable site, optionally using the unconstrained age estimates from the Inside-Outside algorithm of `tsdate`. The second function is `add_sites_time`: a wrapper for a few `tsinfer` and `tsdate` functions which adds ages to the SampleData.sites_time Zarr array. 

Also adds a tutorial which shows how to perform such inference.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -100,7 +100,72 @@ large sections of the tree which do not have any variable sites using
 
 .. _troubleshooting:
 
+*******************************************************************
+Inferring and Dating Tree Sequences with Historic (Ancient) Samples
+*******************************************************************
 
+``tsdate`` and ``tsinfer`` can be used together to infer tree sequences from both
+modern and historic samples. The following recipe shows how this is accomplished
+with a few lines of Python. The only requirement is a tsinfer.SampleData file with
+modern and historic samples (the latter are specified using the `individuals_time`
+array in a tsinfer.SampleData file).
+
+.. code-block:: python
+
+   import msprime
+   import tsdate
+   import tsinfer
+   import tskit
+
+   import numpy as np
+
+
+   def make_historical_samples():
+       samples = [
+           msprime.Sample(population=0, time=0),
+           msprime.Sample(0, 0),
+           msprime.Sample(0, 0),
+           msprime.Sample(0, 0),
+           msprime.Sample(0, 1.0),
+           msprime.Sample(0, 1.0)
+       ]
+       sim = msprime.simulate(samples=samples, mutation_rate=1, length=100)
+       # Get the SampleData file from the simulated tree sequence
+       # Retain the individuals times and ignore the sites times.
+       samples = tsinfer.SampleData.from_tree_sequence(
+         sim, use_sites_time=False, use_individuals_time=True)
+       return samples
+
+   def infer_historic_ts(samples, Ne=1, mutation_rate=1):
+      """
+      Input is tsinfer.SampleData file with modern and historic samples.
+      """
+      modern_samples = samples.subset(np.where(samples.individuals_time[:] == 0)[0])
+      inferred_ts = tsinfer.infer(modern_samples) # Infer tree seq from modern samples
+      # Removes unary nodes (currently required in tsdate), keeps historic-only sites
+      inferred_ts = inferred_ts.simplify(filter_sites=False) 
+      dated_ts = tsdate.date(inferred_ts, Ne=Ne, mutation_rate=mutation_rate) # Date tree seq
+      sites_time = tsdate.sites_time_from_ts(dated_ts)  # Get tsdate site age estimates
+      dated_samples = tsdate.add_sampledata_times(
+         samples, sites_time) # Get SampleData file with time estimates assigned to sites
+      ancestors = tsinfer.generate_ancestors(dated_samples)
+      ancestors_w_proxy = ancestors.insert_proxy_samples(
+         dated_samples, allow_mutation=True)
+      ancestors_ts = tsinfer.match_ancestors(dated_samples, ancestors_w_proxy)
+      return tsinfer.match_samples(
+         dated_samples, ancestors_ts, force_sample_times=True) 
+
+   samples = make_historical_samples()
+   inferred_ts = infer_historic_ts(samples)
+   
+We simulate a tree sequence with six sample chromosomes, four modern and
+two historic. We then infer and date a tree sequence using only the modern
+samples. Next, we find derived alleles which are carried by the historic samples and use
+the age of the historic samples to constrain the ages of these alleles. Finally, we
+reinfer the tree sequence, using the date estimates from tsdate and the historic 
+constraints rather than the frequency of the alleles to order mutations in ``tsinfer``.
+Historic samples are added to the ancestors tree sequence as `*proxy nodes*, in addition
+to being used as samples <https://tsinfer.readthedocs.io/en/latest/api.html?highlight=proxy#tsinfer.AncestorData.insert_proxy_samples>`_.
 
 ++++++++++++++++++++++++++++++
 Command Line Interface Example
--- a/tsdate/date.py
+++ b/tsdate/date.py
@@ -677,7 +677,7 @@ class InOutAlgorithms:
         self.outside = outside
         return posterior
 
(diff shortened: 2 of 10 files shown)


tsdate/util.py
```diff
@@ -121,7 +124,104 @@ def preprocess_ts(tree_sequence, minimum_gap=1000000, trim_telomeres=True):
...
+                try:
+                    nodes_age[index] = json.loads(met.decode())["mn"]
+                except (KeyError, json.decoder.JSONDecodeError):
+                    raise ValueError("Tree Sequence must be tsdated with the\
+                                      Inside-Outside Method. Use unconstrained=False if\
+                                      not.")
+    return nodes_age
+
+
+def sites_time(tree_sequence, unconstrained=True, mutation_age="child",
```
hyanwong, line 149: The function name is a bit baffling here to a newbie, I think. Should we call it "sites_time_from_ts" or something? Also, I'm wondering if "unconstrained", "mutation_age", and "ignore_multiallelic" should be keyword only (but maybe I'm going a bit overboard with the keyword-only args)

tsdate/util.py
```diff
@@ -121,7 +124,104 @@ def preprocess_ts(tree_sequence, minimum_gap=1000000, trim_telomeres=True):
...
+        multiallelic sites. (defaults to True).
+        Default: True
+    :return: An array of length tree_sequence.num_sites giving the estimated site time
+        of each site.
+    :rtype numpy.array
+
+    """
+    if tree_sequence.num_sites < 1:
+        raise ValueError("Invalid tree sequence: no sites present")
+    if mutation_age not in ["arithmetic", "geometric", "child"]:
```
hyanwong, line 174: How about "parent" - have we tested this in the unit tests?

tsdate/util.py
```diff
@@ -121,7 +124,104 @@ def preprocess_ts(tree_sequence, minimum_gap=1000000, trim_telomeres=True):
...
+
+
+def add_sampledata_times(samples, sites_time):
+    """
+    Return a tsinfer.SampleData file with the given times associated with sites. Also
+    checks that each sites' time is at least as old as the oldest historic sample
+    carrying a derived allele at that site.
+
+    :param tsinfer.formats.SampleData samples: A tsinfer SampleData object to
+        add site times to. If both a SampleData file is provided and constrain_historic
```
hyanwong, line 213: Where is "constrain_historic"?

tsdate/util.py
```diff
@@ -121,7 +124,104 @@ def preprocess_ts(tree_sequence, minimum_gap=1000000, trim_telomeres=True):
...
+        is True, use any historic individuals in this SampleData file to constrain site
+        times.
+    :return A tsinfer.SampleData file with sites_times
```
hyanwong, line 216: Don't all SampleData feels have sites_times? So this needs a little more explanation

tsdate/util.py
```diff
@@ -121,7 +124,104 @@ def preprocess_ts(tree_sequence, minimum_gap=1000000, trim_telomeres=True):
...
+                    sites_time[site.id] = age
+            if len(site.mutations) > 1 and sites_time[site.id] == 0:
+                sites_time[site.id] = eps
+            if ignore_multiallelic is False:
+                if len(alleles) > 2:
+                    sites_time[site.id] = tskit.UNKNOWN_TIME
+    return sites_time
+
+
+def add_sampledata_times(samples, sites_time):
```
hyanwong, line 206: Do you think this should be in tsinfer instead?

hyanwong (commented):
Seems pretty good @awohns - quite a few comments about syntax & clarification, though, sorry!

docs/tutorial.rst
```diff
@@ -100,7 +100,69 @@ large sections of the tree which do not have any variable sites using
...
+           msprime.Sample(0, 1.0),
+           msprime.Sample(0, 1.0)
+       ]
+       return msprime.simulate(samples=samples, mutation_rate=1, length=100)
+
+   ts = historical_samples_example()  # Simulate tree sequence
+   samples = tsinfer.SampleData.from_tree_sequence(
+      ts, use_sites_time=False) # Get SampleData file
+   modern_samples = samples.subset(np.where(samples.individuals_time[:] == 0)[0])
+   # The following line removes mutations which are only present in ancient samples
```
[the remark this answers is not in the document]
awohns: We don't really need to do this if we change the simplify call to simplify(filter_sites=False)

docs/tutorial.rst
```diff
@@ -100,7 +100,69 @@ large sections of the tree which do not have any variable sites using
...
+
+   import msprime
+   import tsdate
+   import tsinfer
+   import tskit
+
+   import numpy as np
+
+
+   def historical_samples_example():
```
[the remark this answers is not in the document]
awohns: Good idea!
I'll call it `make_historical_samples` to avoid confusion with "getting" an attribute

docs/tutorial.rst
```diff
@@ -100,7 +100,69 @@ large sections of the tree which do not have any variable sites using
...
+   ts = historical_samples_example()  # Simulate tree sequence
+   samples = tsinfer.SampleData.from_tree_sequence(
+      ts, use_sites_time=False) # Get SampleData file
+   modern_samples = samples.subset(np.where(samples.individuals_time[:] == 0)[0])
+   # The following line removes mutations which are only present in ancient samples
+   samples = samples.subset(
+      sites=np.where(
+         np.any(modern_samples.sites_genotypes[:] == 1, axis=1))[0])
+   inferred_ts = tsinfer.infer(modern_samples) # Infer tree seq from modern samples
+   inferred_ts = inferred_ts.simplify() # Remove unary nodes
```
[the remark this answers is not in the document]
awohns: Good point.
Also, this is the key line I'll amend to include ancient mutations (adding `filter_sites=False`).

tsdate/date.py
```diff
@@ -677,7 +677,7 @@ def outside_pass(self, *, normalize=False, progress=None,
         self.outside = outside
         return posterior
 
-    def outside_maximization(self, *, eps=1e-6, progress=None):
+    def outside_maximization(self, Ne, eps, *, progress=None):
```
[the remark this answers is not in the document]
awohns: Sure

docs/tutorial.rst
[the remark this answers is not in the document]
hyanwong: Ah, so it's only to make sure the site IDs are unchanged? I think it would be better to put this in the `simplify` call, which makes it explicit (and also saves a line of code).

docs/tutorial.rst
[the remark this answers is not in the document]
hyanwong: I guess you might also want to put all the other (pipeline) steps into a single function in the example, so that people can simply copy and paste it for their own use?

docs/tutorial.rst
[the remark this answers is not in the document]
awohns: It removes any "ancient-only" mutations so that all the sites will line up when calling `np.maximum()` in `add_sampledata_times()`. But this isn't necessary if we ensure that all the ancient-only sites are also present in the tree sequence inferred from only modern samples. The only thing is that this won't be the case with real data I expect...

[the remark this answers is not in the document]
awohns: But actually if we tell them to come with a sampledata file of moderns and ancients, then it'll work out fine with this example.

tsdate/util.py
awohns: I can add an option for "parent". I've tested the other three in the unit tests

tsdate/util.py
awohns: Good catch! Holdover from previous version of function

tsdate/util.py
awohns: I thought it could be... Jerome wasn't keen on it though

tsdate/util.py
hyanwong: It's only that you mention it in the docstring...

tsdate/util.py
awohns: Sounds good

PR #142, opened 2020-11-23
fix SVG logo

awohns:


hyanwong:
Hmm, looks off to me, @awohns, sorry!: 
image

Have you tried the font I sent you?

awohns:
Strange, this is what it looks like for me. I used Aglet Slab, same font as tskit

awohns:
Strange, this is what it looks like for me. I used Aglet Slab, same font as tskit

Screenshot 2020-11-23 at 1 53 32 PM

hyanwong:
Did you embed the font, in the svg, though? If not, and the font isn't on the user's system, then the SVG library will do a font substitution, which is why I see the bad logo.

Shouldn't make a difference for the pdf, though.

If you are doing it in (say) Inkscape, you can probably convert the characters to a path instead, which would save doing the substitution.

hyanwong:
P.s. isn't the circle of the "d" a bit small? Try putting the actual letter d from that font in there, then overlaying the clock as near to it as possible before deleting the d

PR #143, opened 2020-11-24
Remove unneeded dict in parameter call

hyanwong:


awohns:
This should actually be `inferred_ts = tsdate.preprocess_ts(inferred_ts, filter_sites: False)`

hyanwong:
It has an equals doesn't it?

PR #148, opened 2020-11-25
Changes to sites_time_from_ts

hyanwong:
A fair bit of documentation change, and some other changes as discussed on slack. In particular, the parameter name has changed from mutation_age to node_selection, ignore_multiallelic has been removed, nodes_time now only gets unconstrained times (and the logic is moved into  sites_time_from_ts), and the epsilon value is renamed min_time and set to default to 1 for *all* zero-time sites. This means setting min_time=0 in lots of the unit tests, but I think that's OK.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -25,7 +25,6 @@ Test cases for the python API for tsdate.
 """
 import collections
 import json
-import math
 import unittest
 import warnings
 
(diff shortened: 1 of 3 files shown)


awohns:
Looks great! Doc string in `nodes_time_unconstrained()` has small typo

awohns:
Happy to merge this now @hyanwong ?

hyanwong:
Yep, the tsinfer switch has been made, so merge I reckon.

Interestingly, until https://github.com/tskit-dev/tsinfer/pull/363 is approved, the `sites_time_from_ts` function shouldn't work to create times for non-variable sites in tsinfer at the moment. So that implies to me that we are missing some tests - notably to feed tree seqs with nonvariable sites into `sites_time_from_ts` *and then use them for inference*.

hyanwong:
I'll write a test to check this, I guess.

awohns:
Hi @hyanwong, can we rebase and merge this PR when you get a chance?

hyanwong:
Rebased. Do you want to merge?

PR #156, opened 2020-12-01
Use Unscaled Time Internally

awohns:
Remove "Ne" from likelihood calculations and only use it in the prior. Doesn't affect output at all.
Look good @hyanwong ?

### DIFF
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -27,13 +27,13 @@ class TestSetCacheDir(unittest.TestCase):
         if os.path.isfile(fn):
             raise AssertionError(f"The file {fn} already exists. Delete before testing")
         with self.assertLogs(level="WARNING") as log:
-            priors_approx10 = ConditionalCoalescentTimes(10)
+            priors_approx10 = ConditionalCoalescentTimes(10, Ne=1)
             assert len(log.output) == 1
             assert "user cache" in log.output[0]
         priors_approx10.add(10)
         # Check we have created the prior file
         assert os.path.isfile(fn)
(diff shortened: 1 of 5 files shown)


tsdate/date.py
```diff
@@ -858,12 +858,10 @@ def outside_maximization(self, Ne, *, eps, progress=None):
...
-        return (
-            self.lik.timepoints[np.array(maximized_node_times).astype("int")] * 2 * Ne
-        )
+        return self.lik.timepoints[np.array(maximized_node_times).astype("int")]
 
 
-def posterior_mean_var(ts, timepoints, posterior, Ne, *, fixed_node_set=None):
+def posterior_mean_var(ts, timepoints, posterior, *, fixed_node_set=None):
     """
     Mean and variance of node age in scaled time. Fixed nodes will be given a mean
```
hyanwong, line 866: This is not in scaled time any more is it?

hyanwong:
Looks great to me. Do you agree that it makes the algorithm easier to understand / more logical?

hyanwong:
We should adjust the docs for the main `date()` method to state that `Ne` is (now) only used to make the prior, so if you provide your own prior, the `Ne` parameter is ignored.

hyanwong:
Can this be merged now?

awohns:
Not quite yet, I need to write a few more tests then will merge. Thanks for checking in on it though, fell off my radar

swamidass:
Is this going to be merged at any point soon?

awohns:
Yep, will merge this week @swamidass

swamidass:
Okay. I need this for that Ne implementation.

PR #171, opened 2021-03-24
Address underflow in log-sum-exp

awohns:
The current implementation "log-sum-exp" in the `LogLikelihoods` class caused underflow in extreme cases on a few human chromosomes. This is because we simply bring the log values into linear space to sum them: the naive implementation of log-sum-exp.
This PR simply changes to using the "streaming'" version of log-sum-exp that was also implemented but unused

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -44,7 +44,6 @@ from tsdate.date import get_dates
 from tsdate.date import InOutAlgorithms
 from tsdate.date import Likelihoods
 from tsdate.date import LogLikelihoods
-from tsdate.date import LogLikelihoodsStreaming
 from tsdate.date import posterior_mean_var
 from tsdate.prior import ConditionalCoalescentTimes
 from tsdate.prior import fill_priors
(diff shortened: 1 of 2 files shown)


hyanwong:
ISTR that we didn't use it because it was a bit slower, but stability should trump speed, so just merge this I think.

hyanwong:
Oh, actually, perhaps add a quick unit test that checks it does the right thing with with `-np.inf` / `0` etc?

awohns:
Hmm, I think the check should be, does it do the right thing with very small values that would underflow in linear space but are captured in log space. i.e.:
```
import numpy as np

X = [-800, -801]
r = 0
for x in X:
    r += np.exp(x) # underflow so that -746 term goes to 0 but -745 is captured

underflowed_answer = np.log(r) 

def logsumexp(X):
        alpha = -np.Inf
        r = 0.0
        for x in X:
            if x != -np.Inf:
                if x <= alpha:
                    r += np.exp(x - alpha)
                else:
                    r *= np.exp(alpha - x)
                    r += 1.0
                    alpha = x
        return np.log(r) + alpha

correct_answer = logsumexp(X)

```

hyanwong:
Yes, that would be useful too.

awohns:
Not sure I'm following what you mean

hyanwong:
In the streaming code there's a line that says `if x != -np.Inf:`, so we should check that it works when `x==-np.Inf`

awohns:
Thanks @hyanwong, I've written two tests to address these points

hyanwong:
Great. Merge it!

PR #172, opened 2021-04-11
Use num_edges directly

hyanwong:
For some reason we use `len(self.ts.tables.nodes.time[self.ts.tables.edges.child[:]])` rather than simply self.ts.num_edges


hyanwong:
I've just added some extra fixes so that we are using the correct types when iterating over edges

awohns:
Good catch

PR #184, opened 2021-06-30
update CHANGELOG

awohns:


CHANGELOG.rst
```diff
@@ -1,3 +1,20 @@
+--------------------
+[0.1.4] - 2021-06-30
+--------------------
+
+**Features**
+
+- The algorithm now operates completely in unscaled time (in units of generations) under
+  the hood, which means that ``tsdate.build_prior_grid`` now requires the parameter
```
benjeffery, line 8: This sounds like a breaking change, which we normally list under a separate heading at the top so it is obvious. It's not a big deal as we're 0.X but worth doing in future.

awohns: Thanks @benjeffery! I changed this for the release notes but didn't do it for the changelog... will keep in mind for the future

PR #188, opened 2021-11-10
require mutation_rate, fail if unsimplified

awohns:
To prevent users from inadvertently running `tsdate` without a mutation rate or using an unsimplified tree sequence from `tsinfer`, this PR makes `mutation_rate` are required parameter and fails if the tree sequence is unsimplified

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -159,11 +159,8 @@ class TestNodeTipWeights(unittest.TestCase):
         ts = utility_functions.two_tree_ts().keep_intervals([(0, 0.2)], simplify=False)
         n = ts.num_samples
         # Here we have no reference in the trees to node 5
-        with self.assertLogs(level="WARNING") as log:
+        with pytest.raises(ValueError, match="nodes not in any tree"):
             SpansBySamples(ts)
-            assert len(log.output) > 0
-            assert "5" in log.output[-1]  # Should mention the node number
-            assert "simplify" in log.output[-1]  # Should advise to simplify
         ts = ts.simplify()
         span_data = self.verify_weights(ts)
         # Root on (deleted) R tree is missing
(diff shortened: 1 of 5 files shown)


awohns:
@benjeffery, I'm getting an error I haven't seen before:
```
attr.exceptions.NotAnAttrsClassError: <class 'tskit.tables.EdgeTableRow'> is not an attrs-decorated class.
```
Do you know what could be causing this?

jeromekelleher:
We're not using attrs in tskit any more @awohns (it wasn't part of the public API) and it looks like you are depending on table rows being attrs instances?

benjeffery:
We changed them to be dataclasses, most of the functionality is the same, but it is in the python standard library.

tests/test_functions.py
```diff
@@ -1883,6 +1875,7 @@ def historical_samples_example(self):
             samples=samples, mutation_rate=1, length=1e2, random_seed=12
         )
 
+    @pytest.mark.skip("Now that we don't allow unary nodes, this example no longer works")
```
hyanwong, line 1878: I don't understand why this is a problem with unary nodes? I think we just need to simplify the inferred_ts, don't we?

hyanwong:
@awohns do you want to rebase this and fix the conflicts? I think the main branch now has a slightly better solution, using `append`, but all the other changes in this PR are worth keeping.

awohns:
@hyanwong, looks like there's some issue with the way that a tree sequence with historical samples is handled...

> ValueError: The input tree sequence has unary nodes: tsdate currently requires that these are removed using `simplify(keep_unary=False)`

I think this happens because we [simplify the tree sequence to only the contemporaneous samples here.](https://github.com/tskit-dev/tsdate/blob/12ee9a2a06be15f4051d9e1ff03a7bb29532a188/tsdate/prior.py#L1045).

Do you think we can just remove that?

hyanwong:
Hi Wilder. Sorry for the delay. Don't we just do the following in `test_historical_samples()`? 

```
inferred_ts = tsinfer.infer(modern_samples, simplify=True).simplify()
```

hyanwong:
Just tested it. Here's what you need:

Screenshot 2022-01-27 at 17 33 45

PR #192, opened 2022-01-07
Use new append functionality

hyanwong:
(edges are not longer an attrs instance)

This allows tests to pass again


awohns:
I guess bringing this up to date with #193 will fix this issue?

hyanwong:
> I guess bringing this up to date with #193 will fix this issue?

Should do, but there are some other fixes needed, so let me rebase etc first

Issue #190, opened 2021-11-23
Sampling times from posterior


awohns:
A number of applications (including [importance sampling](https://www.biorxiv.org/content/10.1101/2021.07.13.452277v1)) would benefit from sampling from the `tsdate` posterior. While this is currently possible with the [mean and variance stored as metadata in tree sequences dated with the inside-outside algorithm](https://github.com/tskit-dev/tsdate/blob/2c200bcf32f2a57c355fba45d770d280bb482982/tsdate/core.py#L886), we should add a function to make this easy for users.

hyanwong:
We can return the posteriors with the (undocumented) [get_dates](https://github.com/tskit-dev/tsdate/blob/9d5d5a497c70dcbea1edbe3f89589dab6c6bec7f/tsdate/date.py#L1016) function. Since users seems to want to look at the posterior, I suggest that we might want to expose this somehow. We could either do this by documenting `get_dates()`, or (perhaps more efficiently) adding a flag such as `return_posterior` to the `tsdate.date` function, which would change the returned value to a tuple of `(tree_seq, posteriors)`, much like the `map_nodes` parameter changes the return value of `ts.simplify()`.

What do we think, @jeromekelleher  and @awohns ? Doing this would avoid some of the issues with undocumented functionality that caused sub-optimal use of tsdate e.g. in Rasmus's analysis.

jeromekelleher:
The ``get_dates`` function doesn't really communicate to me that the posterior is being returned. I think adding adding a flag to the top level ``date`` function is probably a good idea like you suggest @hyanwong.

hyanwong:
When we return the posterior, we return a large matrix of probabilities, whose rows should each sum to one. However, we probably want both row and column names for this matrix, with the rows being the node IDs and the columns being labelled by the timeslice (presumably either the bottom or top of the timeslice). Since numpy doesn't allow row and column labels, we could either return a pandas data frame (and hence impose a dependency on pandas) or return a pair of 1D numpy arrays giving the row and column names. Which do we prefer?

jeromekelleher:
Pandas dependency probably isn't terrible here? Or could return a dictionary of column name -> array values which can easily be turned into a pandas dataframe? The start_time for each interval could be one of those columns.

jeromekelleher:
So the idea is we'd have something like this?
```pythno
dated_ts, posterior = tsdate.date(ts, posterior=True) # or whatever the arg name is
[...]

PR #194, opened 2022-01-07
Modify tables

hyanwong:
At the moment in tsdate we do a lot of operations where we turn everything into a tree sequence, then back into tables for editing, then back into a tree sequence, etc. It seems simpler and more efficient to keep everything as a table collection while we are simplifying, adding provenance, etc, then exporting the tree sequence right at the end.

This should also make addressing #190 easier.


awohns:
LGTM!

Issue #189, opened 2021-11-17
Support time units in tree sequence


awohns:
`tskit` version 0.4.0 includes a [metadata field for the time units used in a tree sequence](https://github.com/tskit-dev/tskit/commit/57eb243582f3ea349800b4f37b06572a857a6da3). We need to support this when it is released: possibly the only thing we need to do for now is set the metadata field to generations in the outputted tree sequence.

hyanwong:
We should probably just leave it as "undefined" for the time being, as we don't know if it's generations or years, or thousands of years, or quite what (it depends on the units of the mutation rate).

Later we could have an optional parameter to specify the units, I guess?

hyanwong:
Fixed in #195

PR #195, opened 2022-01-10
Return posterior and set time_units

hyanwong:
Fixes https://github.com/tskit-dev/tsdate/issues/190 and  https://github.com/tskit-dev/tsdate/issues/189


awohns (approved):
All looks great to me!

PR #201, opened 2022-03-04
Default to not filtering sites, pops, indivs in preprocess

### LINKED ISSUE #178: Change preprocess_ts() defaults
[`tsinfer` issue 491](https://github.com/tskit-dev/tsinfer/issues/491) identified a downstream issue caused when`preprocess_ts` removes sites.
Right now `preprocess_ts` [outputs a warning](https://github.com/tskit-dev/tsdate/blob/af8bcd58fde1b95ff1d7d9f7bdc3f111b51af37a/tsdate/util.py#L138) if the number of sites changes, which says:

>"Different number of sites after preprocessing. Try using **{'filter_sites:' False} to avoid this"

Do we actually want `filter_sites=False` as a default? Interested to hear your opinion @hyanwong

hyanwong:
Fixes #178

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,14 @@
 --------------------
+[0.1.6] - ****-**-**
+--------------------
+
+**Breaking changes**
+
+- The standalone ``preprocess_ts`` function now defaults to not removing unreferenced
+  individuals, populations, or sites, aiming to change the tree sequence tables as
+  little as possible.
+
+--------------------
 [0.1.5] - 2022-06-07
 --------------------
 
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -26,7 +26,6 @@ Test cases for the python API for tsdate.
 import collections
 import json
 import unittest
-import warnings
 
 import msprime
 import numpy as np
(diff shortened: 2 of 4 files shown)


hyanwong:
I think this is a no-brainer. Could you check I'm not missing something @awohns. I think preprocessing shouldn't remove any sites by default. Sites in regions that have been trimmed will not be associated with any edges, so simply won't be used by tsdate (which is correct).

hyanwong:
I think I should just merge this, but it would be useful to get a quick look-over, either by @awohns or @benjeffery perhaps? I don't think it's very controversial: tsdate should only touch the node times, and not delete other stuff from the ts, even if that includes unreferenced sites, populations, etc.

benjeffery (approved):
Seems to make sense to me - although I note that it is no longer possible to get the default `simplify` behaviour with this change. That's fine as it stands as it is a boolean flag - but if simplify changed it would be a problem. Maybe the default here should be `False`, such that passing `None` still gets the default behaviour?

hyanwong:
Oh yes, that's a good call. I know that we usually set `None` as the default, but I think your consideration overrides that. Thanks for the suggestion.

PR #212, opened 2022-07-12
Allow unary nodes (not default) when creating prior

hyanwong:
It is sometimes useful to make priors etc with unary nodes, even if it's not the best way to run tsdate. This should allow the user to override the current behaviour of always raising an error


awohns:
LGTM! Thanks @hyanwong!

PR #220, opened 2022-09-09
Accuracy test

### LINKED ISSUE #219: Add a simple / trivial QC method to the test suite?
I wonder whether it's worth adding in a very quick and simple test for tsdate accuracy when performed on a known (simulated) tree sequence of (say) one tree, two trees, and (say) 10 trees. We could simply check that the r-squared of the true versus inferred node times (in log space, I suspect), was at least equal to the current value. I guess we would store some example tree sequences in `tests/data` and run tsdate on them

```
for filename, r_sq_ts, r_sq_post in (
    # r_sq values obtained by running tsdate 0.12 on these static tree seqs
    ["basic_1_tree", 0.94776615238, 0.94776615238],
    ["basic_2_tree", 0.96605244, 0.96605244],
    ["basic_25_tree", 0.92646, 0.92646],
):
    tskit.load(f"data/{filename}.trees")
    dts, posteriors = tsdate.date(ts, Ne=Ne, mutation_rate=mu, return_posteriors=True)
    nonsample_nodes = np.ones(ts.num_nodes, dtype=bool)
    nonsample_nodes[ts.samples()] = False

    # Test the ts times (before constraining for topology)

    r_sq = np.corrcoef(np.log(ts.nodes_time[nonsample_nodes]), np.log(dts.nodes_time[nonsample_nodes]))[0, 1]**2
    assert r_sq >= r_sq_ts  

    # Test the posterior means too.

    post_mean = np.array([
        np.sum(posteriors[i] * posteriors["start_time"]) / np.sum(posteriors[i])
        for i in np.where(nonsample_nodes)[0]
    ])
    r_sq = np.corrcoef(np.log(ts.nodes_time[nonsample_nodes]), np.log(post_mean))[0, 1]**2
    assert r_sq >= r_sq_post
```

hyanwong:
Fixes #219. I'm pretty sure this can be merged as-is @awohns


awohns:
Super, thanks @hyanwong!

PR #224, opened 2022-12-08
Describe the actual time values

hyanwong:
Prompted by a question from @gregorgorjanc


gregorgorjanc:
@hyanwong this is very rapid and useful addition! Thanks!!!! Very clear now.

PR #225, opened 2022-12-16
Fix nested loop var and remove duplicate class name

hyanwong:


hyanwong:
This has no API implications and all tests pass, so I'm merging. Hope that's OK @awohns !

PR #226, opened 2022-12-16
Don't warn when we expect np.log(0) to occur

hyanwong:


hyanwong:
Merging, trivial change to tests only

PR #232, opened 2023-01-08
Fix unit test and posteriors

hyanwong:
Previously we claimed to return posteriors that were probabilities but the testing was broken, so we didn't notice that we weren't scaling the posteriors to add up to one (i.e. they weren't actually probabilities). This fixes the problem, but therefore changes the posteriors that are output. I think this is the right thing to do.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,12 @@
   individuals, populations, or sites, aiming to change the tree sequence tables as
   little as possible.
 
+**Bugfixes**
+
+- The returned posteriors when ``return_posteriors=True`` now return actual
+  probabilities (scaled so that they sum to one) rather than normalised
+  probabilites whose maximum value is one.
+
 --------------------
 [0.1.5] - 2022-06-07
 --------------------
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -37,7 +37,7 @@ import tskit
 import utility_functions
 
 import tsdate
-from tsdate.base import NodeGridValues
+from tsdate import base
 from tsdate.core import constrain_ages_topo
 from tsdate.core import date
 from tsdate.core import get_dates
@@ -797,14 +797,14 @@ class TestNodeGridValuesClass:
         num_nodes = 5
         ids = np.array([3, 4])
         timepoints = np.array(range(10))
-        store = NodeGridValues(num_nodes, ids, timepoints, fill_value=6)
+        store = base.NodeGridValues(num_nodes, ids, timepoints, fill_value=6)
         assert store.grid_data.shape == (len(ids), len(timepoints))
         assert len(store.fixed_data) == (num_nodes - len(ids))
         assert np.all(store.grid_data == 6)
         assert np.all(store.fixed_data == 6)
 
         ids = np.array([3, 4], dtype=np.int32)
-        store = NodeGridValues(num_nodes, ids, timepoints, fill_value=5)
+        store = base.NodeGridValues(num_nodes, ids, timepoints, fill_value=5)
         assert store.grid_data.shape == (len(ids), len(timepoints))
         assert len(store.fixed_data) == num_nodes - len(ids)
         assert np.all(store.fixed_data == 5)
@@ -815,7 +815,7 @@ class TestNodeGridValuesClass:
         fill = {}
         for ids in ([3, 4], []):
             np.random.seed(1)
-            store = NodeGridValues(
+            store = base.NodeGridValues(
                 num_nodes, np.array(ids, dtype=np.int32), np.array(range(grid_size))
             )
             for i in range(num_nodes):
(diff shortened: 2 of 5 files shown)


hyanwong:
I coded it up so that the `inside_outside` function returns the actual posterior value from the method, rather than standardising the values into probabilities (that sum to one), as this might be useful for checking the algorithm, e.g. to see how much the total value changes if we don't normalise.

The conversion to probabilities is then done immediately *after* the posteriors are returned, in the `get_dates()` function.

tsdate/core.py
```diff
@@ -856,12 +856,12 @@ def outside_maximization(self, *, eps, progress=None):
 
 def posterior_mean_var(ts, posterior, *, fixed_node_set=None):
     """
-    Mean and variance of node age in unscaled time. Fixed nodes will be given a mean
+    Mean and variance of node age. Fixed nodes will be given a mean
     of their exact time in the tree sequence, and zero variance (as long as they are
     identified by the fixed_node_set
```
nspope, line 861:
```suggestion
    identified by the fixed_node_set).
```

tsdate/core.py
```diff
@@ -1054,15 +1060,17 @@ def get_dates(
...
     Infer dates for the nodes in a tree sequence, returning an array of inferred dates
-    for nodes, plus other variables such as the distribution of posterior probabilities
+    for nodes, plus other variables such as the posteriors object
     etc. Parameters are identical to the date() method, which calls this method, then
     injects the resulting date estimates into the tree sequence
 
-    :return: tuple(mn_post, posterior, timepoints, eps, nodes_to_date)
+    :return: a tuple of (mn_post, posteriors, timepoints, eps, nodes_to_date).
+        If the "inside_outside" method is used, posteriors will contain the posterior
+        probabilities, else the returned variable will be ``None``.
```
nspope, line 1073:
```suggestion
        probabilities for each node in each time slice, else the returned variable will be ``None``.
```

CHANGELOG.rst
```diff
@@ -8,6 +8,12 @@
   individuals, populations, or sites, aiming to change the tree sequence tables as
   little as possible.
 
+**Bugfixes**
+
+- The returned posteriors when ``return_posteriors=True`` now return actual
+  probabilities (scaled so that they sum to one) rather than normalised
```
nspope, line 14: The use of "normalised probabilities" is a bit confusing here, because normalizing a probability distribution typically means scaling such that it sums/integrates to 1; not the sort of under/overflow-protection that is happening here

tests/test_functions.py
```diff
@@ -876,19 +880,43 @@ def test_clone(self):
...
+        orig = base.NodeGridValues(num_nodes, np.array(ids), np.array([0, 1.2]), 1)
+        orig[make_nan_row][0] = np.nan
+        assert np.all(np.isnan(orig[make_nan_row]) == [True, False])
+        orig.force_probability_space(base.LIN)
+        orig.to_probabilities()
+        for n in orig.nonfixed_nodes:
+            if n == make_nan_row:
+                assert np.all(np.isnan(orig[n]))
+            else:
+                assert np.allclose(np.sum(orig[n]), 1)
```
nspope, line 909: also assert that they're all nonnegative

nspope (approved):
LGTM! Only a small semantic quibble regarding "normalised" (I'd try to avoid using as an adjective in the docs when talking about anything that's rescaled but doesn't sum/integrate to 1), and a question regarding the overflow protection.

hyanwong:
Thanks a lot. I agree about "normalisation" (I'm not sure I named that one). I think "standardised" would be better. I reckon I should change the name of the method to that.

tsdate/base.py
```diff
@@ -130,7 +129,10 @@ def force_probability_space(self, probability_space):
 
     def normalize(self):
         """
-        normalize grid and fixed data so the max is one
+        normalize grid data so the max is one (in linear space) or zero
+        (in logarithmic space)
+
+        TODO - is it clear why we omit the first element of the
```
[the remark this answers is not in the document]
hyanwong: I think it *should* be, but actually we took the timepoints to be the probabilities associated with a specific (discretised) timepoint, and therefore often ignored the zeroth value, as it was assumed that no internal nodes were at time 0.

tsdate/core.py
hyanwong: Thanks a lot. I had to edit a rebase, so I'll make those changes separately.

hyanwong:
Merging. Will tackle renaming "normalise" to "standardize" in a followup PR. Thanks a lot for the feedback @nspope

tsdate/base.py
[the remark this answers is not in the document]
hyanwong: Yes, it was an approximation, I think.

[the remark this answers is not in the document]
hyanwong: Probably worth chatting it over with @awohns

Issue #230, opened 2023-01-08
Returned posteriors incorrect for mutation_rate=None


hyanwong:
For example:

```python
import tsdate
import tskit
tables = tskit.Tree.generate_balanced(2).tree_sequence.dump_tables()
ts = tables.tree_sequence()
grid = tsdate.build_prior_grid(ts, Ne=40, prior_distribution="gamma", timepoints=2000)
ts, posteriors = tsdate.date(ts, mutation_rate=None, priors=grid, return_posteriors=True)
print(posteriors[2])
```

All the posterior probabilities are one in this case. I think that's wrong

hyanwong:
Actually, this is not a bug in the `mutation_rate=None` path, it only when we date a 2 tip tree. The following code with a 3 tip tree seems to work fine.

```python
import tsdate
import tskit
tables = tskit.Tree.generate_comb(3).tree_sequence.dump_tables()
ts = tables.tree_sequence()
display(ts.draw_svg())
grid = tsdate.build_prior_grid(ts, Ne=40, prior_distribution="gamma", timepoints=20)
ts, posteriors = tsdate.date(ts, mutation_rate=0.1, priors=grid, return_posteriors=True)
print(posteriors[ts.first().root])
```

hyanwong:
I think this is the same bug that we identified previously, although I can't find the issue just now. The problem is that the priors do not properly account for tips at time 0. Here's a trivial example:

```python
import tsdate
import tskit
ts = tskit.Tree.generate_comb(2).tree_sequence
grid = tsdate.build_prior_grid(ts, Ne=40, prior_distribution="gamma", timepoints=20)
print(grid[2]) # Gives rray([0., 1., 1., ..., 1., 1., 1.]) rather than a proper distribution
```

hyanwong:
In might be the same issue being picked up in https://github.com/hyanwong/molecular-sample-dating/blob/main/notebooks/uniform_prior.ipynb

hyanwong:
Another place that we ignore the zeroth timeslice is when normalising: https://github.com/tskit-dev/tsdate/blob/e9bd0982fc4a6cbe9afad8f3a867cb7b8bcc37cb/tsdate/base.py#L135

nspope:
[...]

PR #233, opened 2023-01-09
Better timeslices

### LINKED ISSUE #7: Extend the top end of the adaptive time grid to include large values
At the moment we do not include the quantile=1 upper value, as (for the beta distribution) this is infinite. I think this might explain why we often truncate the oldest MRCAs to be too recent. There should be a relatively easy way to extend the time grid to incorporate a higher upper bound (probably something to ask Gil about)

hyanwong:
Finally fixes #7 and produces a much nicer looking fit for large tree sequences, but leads to slightly worse performance for tiny tree sequences such as those tested in test_accuracy.py, because of #230. When we fix that, this PR should provide uniformly better performance, I hope


hyanwong:
E.g. look at the oldest points in the flooring before and after plots:

image
image

awohns:
This is super @hyanwong!!
How about some simple metrics like MSE and spearman's to benchmark?

hyanwong:
> This is super @hyanwong!!
> How about some simple metrics like MSE and spearman's to benchmark?

Yes, I could add MSE / Spearmans to the "unit" tests at https://github.com/tskit-dev/tsdate/blob/main/tests/test_accuracy.py I guess. The tests there are just to check that we don't accidentally introduce worse performance.

hyanwong:
I added Spearman's in #236 @awohns (along with a test that we scale correctly with Ne)- do you want to check it looks OK and then I can merge #236 which will give us a decent backbone for testing changes such as in this PR?

hyanwong:
> do you want to check it looks OK

Never mind, I merged it anyway, as it's only in the test suite.

hyanwong:
Annoyingly this does seem to make accuracy worse in simple cases. Perhaps we should think about fixing the prior first, then return to this.

PR #236, opened 2023-01-09
More accuracy tests

hyanwong:


hyanwong:
This simply adds to the tests, so I think it is uncontroversial. Merging now.

Issue #228, opened 2022-12-16
Allow scaling function for Ne


hyanwong:
Prompted by discussion with @gregorgorjanc and @nspope, I wonder whether, if we wish to allow a variable Ne, we can simply allow a *function* to be passed to [build_prior_grid](https://tsdate.readthedocs.io/en/latest/python-api.html#tsdate.build_prior_grid) (code [here](https://github.com/tskit-dev/tsdate/blob/77aa23074df155dad6cce8be59d4030bf93536fa/tsdate/prior.py#L999)) as an alternative to a fixed Ne parameter.

The way I would imagine it is that we remove the `Ne` from the `ConditionalCoalescentTimes` class, so that the means and variances stored in `base_priors` are quoted in "canonical" coalescent scaled time, and the mixture prior values will also be in canonical time. We then recode `fill_priors` to to the time-rescaling using either a fixed Ne or a function, and allocate timepoints within that function as appropriate.

hyanwong:
At the same time we might as well fix #202

hyanwong:
As @nspope points out in slack, essentially all we are doing here is rescaling the boundaries of the timeslices (i.e. the timepoints), which we could do by having a scaling function `rescale_to_generations(timepoints_in_coalescent_units, ne_func)` that created a new set out timepoints. However, it might be useful to be able to change the number and distribution of timepoints as well, which would entail recalculating the grid.

Assuming that we want to re-use the conditional coalescent prior (see https://github.com/tskit-dev/tsdate/issues/237#issue-1528674616, suggestion (b)), we probably want to store the alpha and beta parameters for the mixture prior somewhere, so that we can reuse them to allocate Ne-scaled priors into the grid. That means we want to store the result of `get_mixture_prior_params` in some sort of cached object.

nspope:
Could be closed with #252, as the `MixturePrior` class saves the prior params and does rescaling

PR #238, opened 2023-01-11
Separate out the grid creating and filling step

hyanwong:
Starts to tackle #228. Ping @nspope . We still need to abstract away the Ne from most of the prior functions and put it into `fill_discretised_grid` (also allowing it to be a function).

I haven't quite thought about how to deal with the timeslices properly yet, either.

### DIFF
--- a/tsdate/prior.py
+++ b/tsdate/prior.py
@@ -46,6 +46,28 @@ class PriorParams(namedtuple("PriorParamsBase", "alpha, beta, mean, var")):
         return np.where([f == fieldname for f in cls._fields])[0][0]
 
 
+class PriorParamsArray(np.ndarray):
+    """
+    Store alpha and beta parameters in an array (used for a set of nodes). Implementation
+    taken from the RealisticInfoArray example in the numpy docs
+    """
+
+    def __new__(cls, input_array, distribution=None):
+        # Input array is an already formed ndarray instance
+        # We first cast to be our class type
+        obj = np.asarray(input_array).view(cls)
+        # add the new attribute to the created instance
+        obj.distribution = distribution
+        # Finally, we must return the newly created object:
+        return obj
+
+    def __array_finalize__(self, obj):
+        # see InfoArray.__array_finalize__ for comments
+        if obj is None:
+            return
+        self.distribution = getattr(obj, "distribution", None)
+
+
 def lognorm_approx(mean, var):
     """
     alpha is mean of underlying normal distribution
@@ -341,7 +363,10 @@ class ConditionalCoalescentTimes:
         seen_mixtures = {}
         # allocate space for params for all nodes, even though we only use nodes_to_date
         num_nodes, num_params = spans_by_samples.ts.num_nodes, len(param_cols)
-        priors = np.full((num_nodes + 1, num_params), np.nan, dtype=base.FLOAT_DTYPE)
+        priors = PriorParamsArray(
+            np.full((num_nodes + 1, num_params), np.nan, dtype=base.FLOAT_DTYPE),
+            distribution=self.prior_distr,
+        )
         for node in tqdm(
             spans_by_samples.nodes_to_date,
             total=len(spans_by_samples.nodes_to_date),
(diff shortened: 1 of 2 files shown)


nspope:
I think what we need is a function `rescale_to_generations_ago(coalescent_time_ago, population_size)`. If `population_size` is a single number, then this just multiplies by 1/Ne. If it's an array (with columns epoch start / population size) then treat as piecewise-constant population size history. If it's a callable, then use quadrature to get the time rescaling.

tsdate/prior.py
```diff
@@ -947,48 +972,55 @@ def gamma_cdf(t_set, alpha, beta):
...
-    one row for each node in the TS (including fixed nodes)
-    and fill out a NodeGridValues object with the prior values from the
-    gamma or lognormal distribution with those parameters.
+    datable_nodes = np.ones(ts.num_nodes, dtype=bool)
+    datable_nodes[ts.samples()] = False
+    datable_nodes = np.where(datable_nodes)[0]
+    return base.NodeGridValues(
+        ts.num_nodes,
+        datable_nodes[np.argsort(ts.tables.nodes.time[datable_nodes])].astype(np.int32),
+        timepoints,
```
nspope, line 986: This is the only place the time rescaling needs to be applied, I believe.

nspope:
> I think what we need is a function `rescale_to_generations_ago(coalescent_time_ago, population_size)`

I can put this together -- what's the best way to put stuff into this draft, @hyanwong? Make a PR at your fork?

hyanwong:
> > I think what we need is a function `rescale_to_generations_ago(coalescent_time_ago, population_size)`
> 
> I can put this together -- will make a separate PR with the function & tests, and we can worry about integrating it later.

That would be great, thanks.

tsdate/prior.py
```diff
@@ -947,48 +972,55 @@ def gamma_cdf(t_set, alpha, beta):
...
-
     # TO DO - this can probably be done in an single numpy step rather than a for loop
     for node in tqdm(
-        datable_nodes, desc="Assign Prior to Each Node", disable=not progress
+        grid.nonfixed_nodes, desc="Assign Prior to Each Node", disable=not progress
     ):
         with np.errstate(divide="ignore", invalid="ignore"):
-            prior_node = cdf_func(timepoints, main_param[node], scale=scale_param[node])
+            prior_node = cdf_func(
+                grid.timepoints, main_param[node], scale=scale_param[node]
```
hyanwong, line 1016: If we quote the params in coalescent scaled time here, we need `grid.timepoints` to be in coalescent time too. It seems to me that we would then need to apply a conversion from coalescent time to "true" time, using whatever function was passed in from the coalescent point distribution

tsdate/prior.py
hyanwong: I think that we should maintain "true" (i.e. generations, years, or whatever) time within the NodeGridValues object, because people might want to specify their own timepoints (in e.g. generations). There is also the issue that with historical samples, we will want to truncate the prior to the oldest fixed node child, where the time of the child is quoted in either years or generations.

I've put a comment in another place where I think we might need to consider the scaling.

hyanwong:
Another point: we almost certainly want to store the `mixture_prior` for reuse, as it's expensive to recalculate all those mixtures. That's why I created a new `PriorParamsArray` to store the results of `priors.get_mixture_prior_params`. As long as those are stored in coalescent scaled time, we can keep them around for reuse on subsequent rounds of iteration.

nspope:
I'm thinking something along the lines of:
```python
def rescale_time_piecewise(time_ago, breakpoints, time_measure):
    """                                                                                                                                               
    Rescales time given a piecewise-constant time measure (e.g. a piecewise
    constant demographic history).  To convert from generations to coalescent
    units, the time measure per epoch should be effective population size.  To
    convert from coalescent units to generations, the time measure should be
    the coalescent rate.

    Returns times and breakpoints under the new measure::

        generations, breaks_in_generations, population_size = rescale_time_piecewise(
            coalescent_units, breaks_in_coalescent_units, coalescent_rate
        )
        coalescent_units, breaks_in_coalescent_units, coalescent_rate = rescale_time_piecewise(
            generations, breaks_in_generations, population_size
        )
    """
    ...
```
For the CDF: the lower tail probabilities (y-axis) stay the same, it's just the quantiles (x-axis) that need to be rescaled. So, given a function like the above, we can just calculate CDF per node in terms of coalescent time units (perhaps converted from user-input grid in generations), then rescale the time grid back to generations at the end.

nspope:
It would be easy to support a user-input continuous population size function; by sampling this function on a fine grid then doing the piecewise time conversion as outlined above. Which is effectively a quadrature scheme.

nspope:
One issue is creating a "default" time discretization when different nodes belong to different populations. Currently, the discretization is done by taking the union of quantiles across the priors for nodes, and thinning these to some minimum distance.

If all nodes have the same population size history, then this'll still work -- because there's just one transformation/time-rescaling induced by the shared population size history. But, if different nodes have different population size histories, it gets more complicated. The "union-of-quantiles" approach would have to be done *after* the time rescaling, which would require setting a minimum distance (in units of generations) for what constitutes a unique timepoint.

hyanwong:
I'm not sure I quite understand the problem with creating the quantiles after time rescaling? I'm assuming that the timeslices for discrete population size changes aren't the same as the timeslices for discretization (and indeed the population size changes could be a smooth function rather than a piecewise one). So as long as we can sum over the individually rescaled node time distributions (which admittedly might be a bit tedious), we can generate a timepoints grid, I think?

hyanwong:
Actually, I think I misunderstood above, sorry. But I still think that it's reasonable to create timepoints after the rescaling has occurred. I suspect you (@nspope ) will have a better idea for how to do this than I do, though!

nspope:
Yes, it's definitely doable, it just requires that the thinning be based on units of generations rather than coalescent time. If I'm understanding correctly [here](https://github.com/tskit-dev/tsdate/blob/4a81aec9cabcb182695caa1ae2375a7b3bb25f89/tsdate/prior.py#L935-L936).

hyanwong:
Ah yes, we would indeed need to switch to points on a timescale that was universal across nodes. As it is, I think we should be storing the timepoints in universal units within the NodeGridValues object anyway, so we have to do the conversion at some point.

I don't know if we would want to reallocate timepoints in each iteration: it seems from what you have said that you think that would be a good idea.

hyanwong:
Replaced with #248

PR #245, opened 2023-01-16
Rename "reduce" to "ratio"

hyanwong:
To reflect that we are actually calculating likelihood ratios here (dividing in linear space).

Any comments on this terminology @nspope or @awohns ?


nspope (approved):
Good call, looks fine to me.

awohns:
Sounds good to me too

PR #248, opened 2023-01-25
Priors with piecewise demographic histories

nspope:
- Change `Ne` argument to `population_size` in the API
- If `population_size` is a 2d array with two columns, treat these as epoch start times and population sizes
- The conditional coalescent prior no longer scales its parameters by Ne; instead, this is done by rescaling the time grid within `fill_priors`
- Supercedes #238 
- It'd be good to benchmark these priors, along the lines of what is done in `evaluation/evaluate_accuracy.py`.

It'd be nice to have this work for arbitrary functions, by using numerical integration to do the time rescaling. This could fail for discontinuous functions, though, so there would have to be some convergence checks.

It'd also be nice to be able to specify different demographic histories for different nodes, but there's some issues to sort out with how the time grid is constructed.


nspope:
Seems to work OK -- here's a couple simple test cases, written as "epoch start (population size)".  

- "Bump": 0 (1e4) ; 10000 (1e5) ; 50000 (1e3) 
- "Expand": 0 (1e5) ; 10000 (1e4) ; 50000 (1e3) 
- "Bottleneck": 0 (1e5) ; 10000 (1e3) ; 50000 (1e4)

with 1 Mb and 100 samples. For a fair "constant Ne" comparison, I used 1/2 the expected pair coalescence rate under the demographic model:
<center>

</center>

The bias with the constant-Ne prior (red points) seems to be quite a bit worse, esp. for expansion.

nspope:
I don't want to cram too much into this PR, but it's worth thinking about how to easily reuse the mixture prior across different Ne/time discretizations. One way to do this is to define a class that is separate from `ConditionalCoalescenceTimes`, that maps the conditional coalescent prior onto a particular tree sequence upon instantiation; stores the prior hyperparameters per node; and creates grids for a given time discretization/population size history.

If this seems like the way to go, then there's likely some additional refactoring/documenting that could happen on top of these changes (for example some of the mixture-prior-related functionality in `ConditionalCoalescenceTimes` could probably be moved into the new class).

hyanwong:
> Seems to work OK -- here's a few simple test cases under recent exponential growth/decline, for simulated (not inferred) 1 Mb tree sequences with 100 samples. For a fair "constant Ne prior" comparison, I used 1/2 the expected pair coalescence rate under the demographic model:
> 
> 
> 
> The bias with the constant-Ne prior (red points) seems to be quite a bit worse for the growing/declining scenarios.

This is fantastic. Thanks a lot for taking this on @nspope . I'll get round to review tomorrow.

hyanwong:
> I don't want to cram too much into this PR, but it's worth thinking about how to easily reuse the mixture prior across different Ne/time discretizations. 

Yes, I completely agree with that.

> One way to do this is to define a class that is separate from `ConditionalCoalescenceTimes`, that maps the conditional coalescent prior onto a particular tree sequence upon instantiation; stores the prior hyperparameters per node; and creates grids for a given time discretization/population size history. (Essentially, refactor `build_prior_grid` into a class).

Yes, I suggest we might want to retain the `ConditionalCoalescentTimes` (which is not dependent upon any particular tree sequence) and the `SpansBySamples` class (which *is* dependent on a particular tree seq), and create a new class that uses both of these and which creates the grid.

> If this seems like the way to go, then there's likely some additional refactoring/documenting that could happen on top of these changes (for example some of the mixture-prior-related functionality in `ConditionalCoalescenceTimes` could probably be moved into the new class).

You are right that the mixture prior calc could be moved out of `ConditionalCoalescenceTimes`, and put somewhere equally good or better!

nspope:
> Yes, I suggest we might want to retain the ConditionalCoalescentTimes (which is not dependent upon any particular tree sequence) and the SpansBySamples class (which is dependent on a particular tree seq), and create a new class that uses both of these and which creates the grid.

Great -- that's exactly what I've done in these last two commits (was planning on waiting on further refactoring, as I'm sure this will require updating a bunch of tests).

hyanwong:
Would it be better for me to review this PR first, or #251 ?

nspope:
This one is bigger and has priority, I think -- thanks @hyanwong.

tsdate/core.py
```diff
@@ -932,7 +932,7 @@ def constrain_ages_topo(ts, post_mn, eps, nodes_to_date=None, progress=False):
 def date(
     tree_sequence,
     mutation_rate,
-    Ne=None,
```
hyanwong, line 935: It would be good to retain `Ne` as a (deprecated) alias for population_size, for backwards compatibility. We can raise a warning that this is deprecated, if you like.

tsdate/prior.py
```diff
@@ -970,10 +973,22 @@ def fill_priors(node_parameters, timepoints, ts, Ne, *, prior_distr, progress=Fa
     datable_nodes = np.ones(ts.num_nodes, dtype=bool)
     datable_nodes[ts.samples()] = False
     datable_nodes = np.where(datable_nodes)[0]
+
+    """
```
hyanwong, line 977: Can we put these as comments (starting with #) rather than triple quoted strings, we we tend to reserve only for docstrings at the start of the function

tsdate/prior.py
```diff
@@ -947,7 +948,9 @@ def gamma_cdf(t_set, alpha, beta):
     return np.insert(t_set, 0, 0)
 
 
-def fill_priors(node_parameters, timepoints, ts, Ne, *, prior_distr, progress=False):
+def fill_priors(
+    node_parameters, timepoints, ts, population_size, *, prior_distr, progress=False
+):
```
hyanwong, line 953: It would be useful to document the fact that the population_size parameter can now be an array (is this actually a strict requirement?)

tsdate/prior.py
```diff
@@ -1011,9 +1139,11 @@ def build_grid(
 
     :param TreeSequence tree_sequence: The input :class:`tskit.TreeSequence`, treated as
         undated.
-    :param float Ne: The estimated (diploid) effective population size: must be
-        specified. Using standard (unscaled) values for ``Ne`` results in a prior where
-        times are measures in generations.
+    :param float population_size: The estimated (diploid) effective population
+        size: must be specified. May be a single value, or a two-column array with
+        epoch breakpoints and effective population sizes. Using standard (unscaled)
```
hyanwong, line 1144: How do the pop sizes map to the breakpoints? Do we require the same number of rows in the 2 columns, and if so, is one ignored?

It would be nice to be able to specify a continuous Ne function here too, instead of a piecewise set of points, wouldn't it? Would this be any technically more difficult as a rescaling method? I don't think it should...

tests/test_functions.py
```diff
@@ -1101,8 +1102,8 @@ def run_outside_algorithm(
         self, ts, prior_distr="lognorm", standardize=False, ignore_oldest_root=False
     ):
         span_data = SpansBySamples(ts)
-        Ne = 0.5
-        priors = ConditionalCoalescentTimes(None, Ne, prior_distr)
+        Ne = np.array([[0, 0.5]])
```
hyanwong, line 1105: I'm not quite understanding why we need to give an array of 2 values here? Is this the timepoint and the Ne? I suspect we could just assume that if a single scalar was given, if could apply to all time.

hyanwong: (also, I wonder if the parameterization should be `[array_of_times, array_of_Ne]` rather than `[[t0, Ne0], [t1, Ne1], ...]`?)

tsdate/prior.py
```diff
@@ -991,16 +1006,129 @@ def fill_priors(node_parameters, timepoints, ts, Ne, *, prior_distr, progress=Fa
...
-    Ne,
+    population_size,
     timepoints=20,
     *,
     approximate_priors=False,
     approx_prior_size=None,
     prior_distribution="lognorm",
-    eps=1e-6,
     # Parameters below undocumented
+    eps=1e-6,  # placeholder
```
hyanwong, line 1131: Nice spot. This was never actually passed through, was it. Doh.

tsdate/util.py
```diff
@@ -300,3 +300,43 @@ def add_sampledata_times(samples, sites_time):
...
+    generations, the time measure should be the coalescent rate ``1/(2 * Ne)``.
+
+    :param np.ndarray time_ago: An increasing vector of time points
+    :param np.ndarray breakpoints: Start times of pieces
+    :param np.ndarray time_measure: Time measure within pieces
+
+    :return: Inputs in new time measure
+    """
+
+    # TODO: Make this work for arbitrary positive fuctions? Could use
```
hyanwong, line 319: "functions" -> "functions"

tsdate/util.py
```diff
@@ -300,3 +300,43 @@ def add_sampledata_times(samples, sites_time):
...
+    # scipy.integrate.quad, but would need to check for convergence;
+    # as it can fail when function is not sufficiently smooth.
+
+    assert sorted(breakpoints)
```
hyanwong, line 323: This doesn't assert that they are sorted, it sorts the array and then asserts the array is not empty.

tsdate/util.py
```diff
@@ -300,3 +300,43 @@ def add_sampledata_times(samples, sites_time):
...
+    :param np.ndarray breakpoints: Start times of pieces
+    :param np.ndarray time_measure: Time measure within pieces
+
+    :return: Inputs in new time measure
+    """
+
+    # TODO: Make this work for arbitrary positive fuctions? Could use
+    # scipy.integrate.quad, but would need to check for convergence;
+    # as it can fail when function is not sufficiently smooth.
+
```
hyanwong, line 322: Should you assert stuff about the relative lengths of the 3 arrays?

tsdate/util.py
```diff
@@ -300,3 +300,43 @@ def add_sampledata_times(samples, sites_time):
...
+
+def change_time_measure(time_ago, breakpoints, time_measure):
+    """
+    Rescales time given a piecewise-constant time measure. To convert from
+    generations to coalescent units, the time measure per epoch should be 2 *
+    effective population size.  To convert from coalescent units to
+    generations, the time measure should be the coalescent rate ``1/(2 * Ne)``.
+
+    :param np.ndarray time_ago: An increasing vector of time points
```
hyanwong, line 312: "Starting at time 0"

tsdate/util.py
```diff
@@ -300,3 +300,43 @@ def add_sampledata_times(samples, sites_time):
...
+    :param np.ndarray breakpoints: Start times of pieces
```
hyanwong, line 313: We require the first breakpoint to be 0, right? So we should probably say this.

tsdate/prior.py
nspope: It does turn out to be technically more difficult, because the user is passing in a continuous population size function $n(t)$ that is parameterized by time in generations. Instead, we need a population size function that is parameterized in coalescent units, because we're converting from coalescent timescale to natural timescale. This won't generally have a closed form. It can be calculated numerically, but it involves optimization where the objective function uses quadrature, and it's pretty easy for this to futz up numerically. (Unless I'm missing something obvious).

On the other hand, if the population size function is piecewise constant, then it's really easy to convert back-and-forth. Ultimately, I think the thing to do (to allow continuous $n(t)$) is to approximate it with piecewise constant, and use a nested quadrature rule (Romberg or something similar) to get the approximation to within a desired accuracy. But, this'll be somewhat involved to implement, and I think it's a good first step to just get the piecewise constant case working first.

tests/test_functions.py
nspope: It's just converting from single scalar to the "piecewise population size" format, which is epoch start (measuring time moving backwards) in the first column, and effective population size in the second column. The conversion can happen later, it'd probably be better to consolidate into the rescaling function itself.

I don't have strong opinions about whether the array should be 2-row or 2-column -- whatever you think is clearer. I do think it should be a 2d numpy array rather than a list of lists, so there's no ambiguity about dimensionality.

tsdate/prior.py
```diff
@@ -970,10 +973,22 @@ def fill_priors(node_parameters, timepoints, ts, Ne, *, prior_distr, progress=Fa
...
+    )
+    generations, _, _ = util.change_time_measure(
+        timepoints, epoch_start, coalescent_rate
+    )
+
     prior_times = base.NodeGridValues(
         ts.num_nodes,
         datable_nodes[np.argsort(ts.tables.nodes.time[datable_nodes])].astype(np.int32),
-        timepoints,
+        generations,
```
hyanwong, line 991: Ultimately we might want to allow times to be specified in years rather than generations (this makes more sense for e.g. cases with historical samples dated to a specific timepoint), and of course is the only natural unit to use when inferring dates from population with overlapping generations, or where average generation time differs between subpopulations. Previously this would have be done by simply submitting rescaled values (e.g. for mutation rate, Ne, etc) to the prior, and changing the `time_units` parameter. I'm a little worried that hard-coding "generations" as the name here fixes us to a generational timescale?

Perhaps all we need, however, is a comment that mentions a "todo" to allow for different timescales to be used.

tsdate/prior.py
```diff
@@ -991,16 +1006,129 @@ def fill_priors(node_parameters, timepoints, ts, Ne, *, prior_distr, progress=Fa
...
+        Calculate prior grid for a set of timepoints and a population size history
+        """
+
+        if isinstance(population_size, np.ndarray):
+            if population_size.ndim != 2:
+                raise ValueError("Array 'population_size' must be two-dimensional")
+            if population_size.shape[1] != 2:
+                raise ValueError(
+                    "Population size array must have two columns that contain \
+                    epoch start times and population sizes, respectively"
```
hyanwong, line 1073: "Epoch start" is in reverse time (i.e. the most recent time in the epoch), so using the word "start" might confuse anyone working with real data who doesn't think in reverse time.

hyanwong (changes requested):
This mostly looks good to me. Thanks a lot for starting with this @nspope. Apart from the few minor suggestions I made yesterday, I just have a few discussion points about the general structure:

1. Is `util.change_time_measure` the right place to put the function? Is there some sort of class we might want to attach it to instead?
2. The use of a 2-d array for timepoints and population sizes seems a bit like cramming 2 different things into a single array. I wonder whether we should either require a list of 2 arrays, of size `n` (population sizes) and `n-1` (timebreaks), which would then remove the naming problem as to whether the timepoints represented the "beginning" or "end" of a population size. Alternatively, we could roll up a class to contain the information.

tsdate/util.py
hyanwong: (Or, as a I suggested in my review, maybe we should have the breakpoints array be of size one less than the population size array?)

tsdate/util.py
hyanwong: I think you want `assert np.all_equal(breakpoints, sorted(breakpoints))`. Or maybe `assert np.all(np.diff(breakpoints) > 0)`?

tests/test_functions.py
hyanwong: I agree about consolidating into the rescaling function. See my other comments about a list of arrays (to allow different lengths for timebreaks and population sizes, meaning that we don't need to require a first "break" at time 0.

tsdate/prior.py
hyanwong: Right, I see. I think this is an integration thing, right? We need the integral of the rate function to convert to successive time points. So I guess we could require the integral to be passed, rather than the actual population size function?

As you say, though, it's good to get the piecewise version working first. I'm always a bit scared of these piecewise things: the discrete jumps between time boundaries feels very wrong to me!

nspope: Hm, I think that'd be a pretty stringent requirement for the user. Except for a few specific cases (piecewise constant, exponential) I don't think the integral would have a closed form. It seems better to support these explictly (e.g. add a "growth_rate" array for piecewise-exponential models) or to do the numerics under the hood by sampling the continous function on successively finer grids until the time rescaling converges. But let's cross that bridge later ...

In my experience, as long as the piecewise constant history is a fine enough discretization, it'll do a very good job of approximating a continuous population size history. After all, a discrete-time Wright-Fisher model is effectively a piecewise-constant one, with each piece the length of a generation.

nspope:
Thanks @hyanwong! To make sure I understand-- when you say use `n` population sizes and `n - 1` timebreaks, you mean that we'd exclude the first break at time 0?

If a class is used to package this up, then the rescaling function could be moved from `util` into this class. Then the ultimate workflow for the user would look something like,
```
# case 1: piecewise constant history
pop_size = PopulationSizeHistory(time_breaks=breaks_array, population_size=population_size_array)
tsdate.date(..., population_size=pop_size)

# case 2: continuous function
# we sample the function on successively finer grids (e.g. nested quadrature scheme) until the time rescaling of a test vector converges
pop_size = PopulationSizeHistory(population_size=population_size_callable)
tsdate.date(..., population_size=pop_size)

# case 3: single scalar, gets transformed to PopulationSizeHistory under the hood
tsdate.date(..., population_size=population_size_scalar)
```

hyanwong:
> Thanks @hyanwong! To make sure I understand-- when you say use `n` population sizes and `n - 1` timebreaks, you mean that we'd exclude the first break at time 0?

Yes. The first population size essentially goes from the current day (i.e. 0) to the first timepoint, etc.

> If a class is used to package up the population size history, then the rescaling function could be moved from `util` into this class. Then the ultimate workflow for the user would look something like,
> 
> ```
> # case 1: piecewise constant history
> pop_size = PopulationSizeHistory(time_breaks=breaks_array, population_size=population_size_array)
> tsdate.date(..., population_size=pop_size)
> 
> # case 2: continuous function
> # we sample the function on successively finer grids (e.g. nested quadrature scheme) until the time rescaling of a test vector converges
> pop_size = PopulationSizeHistory(population_size=population_size_callable)
> tsdate.date(..., population_size=pop_size)
> 
> # case 3: single scalar, gets transformed to PopulationSizeHistory under the hood
> tsdate.date(..., population_size=population_size_scalar)
> ```

That sounds about right to me. For "Population size" we allow either a single number or an object that works in a specific way as required by the estimation method ("quacks like a duck"). That way we can sub in other identically behaving objects too, e.g. an approximation for continuous functions. Then we shouldn't need to accept a "population_size_callable", I think? It can just be an object with the appropriate methods.

I wonder if we should copy the msprime or demes method names somehow? I'm inclined not to, although being able to convert a simple demes spec into the

nspope:
> Then we shouldn't need to accept a "population_size_callable", I think? It can just be an object with the appropriate methods.

For this case, what I'm thinking is that there's a piecewise approximation that happens under the hood when `PopulationSizeHistory` is instantiated with a callable `population_size`; that would make successively finer approximations until the time rescaling for a test grid converged, then store the breakpoints/time measure under both natural and coalescent scales. But maybe we can hash out the details for this API later, I'm a bit hazy on the details. 

I could also see what (I think) you're describing, which is that any object with a `rescale_time` method could be passed. Although this may be more prone to user misspecification.
 
> I wonder if we should copy the msprime or demes method names somehow? I'm inclined not to, although being able to convert a simple demes spec into the population_size_object format might be handy.

I'm also inclined not to-- I think it's nice to keep argument naming consistent for related packages, but unless the classes are 100% interoperable I think it's messy to give them the same name. I agree that a `.from_demes` method would be handy. It might also be handy to have a way to approximate a multipopulation history with a single-population one, by calculating the expected pair coalescence rate under the multipopulation model for all possible pairs of populations, then averaging.

> What do you think @nspope ? Do you have any preferences here?

This all sounds good to me.

hyanwong:
> I could also see what (I think) you're describing, which is that any object with a `rescale_time` method could be passed. Although this may be more prone to user misspecification.

Yes, I think this is the pythonic way to do it. I don't think user misspecification is really an issue. You would have to work hard to make a class of the right sort, in which case caveat emptor.
 
>  I think it's nice to keep argument naming consistent for related packages, but unless the classes are 100% interoperable I think it's messy to give them the same name. I agree that a `.from_demes` method would be handy. 

Great, sounds like we are on the same page.

> It might also be handy to have a way to approximate a multipopulation history with a single-population one, by calculating the expected pair coalescence rate under the multipopulation model for all possible pairs of populations, then averaging.

Yes, I guess we could add approximations like this.

> This all sounds good to me.

:+1:

nspope:
Quick question: Ne is stored in the tree sequence provenance, which is easy enough if it is a scalar. But if variable Ne is stored in a PopulationSizeHistory object, what's the right approach to make it json serializable?

hyanwong:
There isn't really an easy way we have found to do this. I don't think we actually need to store it in the provenance, though. It's the same with a recombination map object in msprime: we simply don't store it in provenance.

It would, perhaps, be useful to note somehow in the provenance that the full details are not available. AFAIK we don't have a way to do this either.

nspope:
Alrighty, this is ready for another look. Shall I update the docs, as well? Or do we want to benchmark this approach, first?

hyanwong (approved):
LGTM. I guess we should merge and improve iteratively. Could you add something to the CHANGELOG? @nspope?

hyanwong:
> Alrighty, this is ready for another look. Shall I update the docs, as well? Or do we want to benchmark this approach, first?

When you say "benchmark", do you mean for speed (I don't think this will be affected) or accuracy of the new approach?

If you are happy to update the docs, please do.

nspope:
OK done! I think I'll postpone updating the docs for a future PR (if that's OK) -- as it'd be nice to first work out how marginal likelihood / Ne optimization fit into this picture.

hyanwong:
Merging, thanks @nspope - this is all great stuff

Issue #249, opened 2023-01-27
Return marginal likelihood from inside pass


nspope:
It'd be useful to return the (approximate) marginal likelihood from the inside pass (that is, the likelihood of the mutations in the tree sequence, integrating over possible node ages). The reason is that with a constant mutation rate, this marginal likelihood should be quite sensitive to the choice of effective population size and time grid. For example, I simulated ten tree sequences from the standard coalescent with population size 26000; and calculated marginal likelihood with tsdate's algorithm across choices of Ne for the prior:

<img src="https://user-images.githubusercontent.com/5767783/214977465-f7a85039-2202-4935-af6d-3d9d445816da.png" width="50%" height="50%">

In this figure, relative absolute error is calculated as `abs(estimated age - true age) / true age` excluding nodes that are less than 100 generations old; because at mu=1e-8 there's not much information with which to estimate these precisely (& they'll dominate the relative error otherwise). The vertical lines are maxima/minima for logLik/RAE for each sim. For this particular case it seems like choosing Ne on the basis of marginal likelihood also does a pretty good job at minimizing the relative absolute error. 

I ran a separate set of simulations where I varied population size between 5000 and 50000, to see how the "inferred" Ne (in terms of maximizing log marginal likelihood) performs as a surrogate for the "optimal" Ne (in terms of minimizing relative absolute error): 

<img src="https://user-images.githubusercontent.com/5767783/214980577-aa352dbd-b16f-40d8-bbe6-244bf20ac208.png" width="50%" height="50%">

There's a downward bias in "optimal" Ne vs the truth regardless of criterion (I think because the current prior discretization scheme biases the expectation of the prior). However, it looks like choosing the prior Ne based on marginal likelihood is a good proxy for minimizing RAE. 

So, this may provide a useful heuristic for "calibrating" the prior in an iterative fashion ala #237, especially when combined with #248 which gives more flexibility for modifying the prior timescale. In particular, if we change the underflow protection scheme so that: 
- Poisson likelihoods aren't standardized over timepoints at each iteration, and
- "Standardization" at each step is done by dividing by the sum, not the max (e.g. use logsumexp when working in logspace)

then the (approximate) marginal likelihood can be calculated directly from the `denominator` member of the inside-outside algorithm class, because geometric scaling appropriately divvies up the contributions of shared descendants across roots.

hyanwong:
[...]

Issue #237, opened 2023-01-11
Iterative approach to variable Ne


hyanwong:
We can attempt to infer a variable Ne by performing a tsdate inference, assessing the coalescence distribution over time (possibly broken up by population, if necessary), then using a rescaled set of timepoints to reflect variable Ne (see #228 ), and performing another tsdate inference. We could continue to iterate like this until the timepoint scaling equilibrates.

There are a couple of issues here. The one that first strikes me is whether in the second round of tsdate usage, we (a) set the prior distribution for each node to be the posterior of the initial inference, or (b) reset the priors to the conditional coalescent but simply rescale the times using the newly assessed Ne mapping. My inclination is to do (b). We know that sometimes the posterior times are a bit off, so using them as a prior seems like we might be feeding Tsdate-related errors back into the system, amplifying any tsdate issues.

I wonder if @awohns and @nspope have any thoughts here?

hyanwong:
> (possibly broken up by population, if necessary)

It may be worth thinking about how we would allocate different Ne transformations for different populations. Since this is simply affecting the prior, we could theoretically allocate a different transformation for each node if we wanted, without working about topological interactions between the nodes.

I think Relate allows separate Ne for separate population (and it's used quite a lot), so it might be worth thinking about how it could be done. Maybe the first thing is to work out how we would assign nodes to populations.

I wonder if it's also worth thinking about what we would do in (semi) continuous space, where we can't clearly assign samples to discrete populations.

hyanwong:
> I wonder if it's also worth thinking about what we would do in (semi) continuous space, where we can't clearly assign samples to discrete populations.

The reason why allocating nodes to populations works is that a population is usually large enough that there are a number of coalescent points within that population during each time slice. In terms of tsdate, we want to allocate a prior distribution to a specific ancestor, and we can use the density of coalescence points within the population in which that ancestor resides as a measure of population size to scale that distribution to fit the demography. In other words, we can average coalescent density among all members of a sub population.

We could try extending that to a continuous population by other sorts of averaging. For instance, we could allocate (rough) geographical locations to ancestors, in the way we did in the unified genealogy paper. We could then have some function which looks at the average coalescent density around a target ancestor, taking all coalescence points from the same time-slice but weighting the density by the inverse squared distance of each coalescence from the target ancestral lineage(s) at that timepoint.

nspope:
[...]

PR #251, opened 2023-01-30
Return marginal likelihood from inside algorithm

nspope:
- Rationale in #249, first step in addressing #237
- `get_dates` has an extra output at the end, which is the marginal likelihood from the inside pass
- This is returned by `date` if `return_likelihood = True`, in a way that's similar to what is done for posteriors
- The only change to the existing algorithm is to keep the Poisson likelihoods unstandardized. This change doesn't seem to increase the risk of over/underflow, given that inside values are standardized at each node.

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -580,14 +580,14 @@ class TestPriorVals:
 
 
 class TestLikelihoodClass:
-    def poisson(self, param, x, standardize=True):
+    def poisson(self, param, x, standardize=False):
         ll = np.exp(-param) * param**x / scipy.special.factorial(x)
         if standardize:
             return ll / np.max(ll)
         else:
             return ll
 
-    def log_poisson(self, param, x, standardize=True):
+    def log_poisson(self, param, x, standardize=False):
         with np.errstate(divide="ignore"):
             ll = np.log(np.exp(-param) * param**x / scipy.special.factorial(x))
         if standardize:
@@ -1012,7 +1012,9 @@ class TestAlgorithmClass:
 
 
 class TestInsideAlgorithm:
-    def run_inside_algorithm(self, ts, prior_distr, standardize=True, **kwargs):
+    def run_inside_algorithm(
+        self, ts, prior_distr, standardize=True, logspace=False, **kwargs
+    ):
         Ne = 0.5
         priors = tsdate.build_prior_grid(
             ts,
@@ -1024,11 +1026,14 @@ class TestInsideAlgorithm:
         )
         eps = 1e-6
         mut_rate = 0.5
-        lls = Likelihoods(ts, priors.timepoints, mut_rate, eps=eps)
+        if logspace:
+            lls = LogLikelihoods(ts, priors.timepoints, mut_rate, eps=eps)
+        else:
+            lls = Likelihoods(ts, priors.timepoints, mut_rate, eps=eps)
         lls.precalculate_mutation_likelihoods()
         algo = InOutAlgorithms(priors, lls)
-        algo.inside_pass(standardize=standardize)
-        return algo, priors
+        marg_lik = algo.inside_pass(standardize=standardize)
+        return algo, priors, marg_lik
 
     def test_one_tree_n2(self):
         ts = utility_functions.single_tree_ts_n2()
@@ -1055,7 +1060,9 @@ class TestInsideAlgorithm:
 
     def test_two_tree_ts(self):
         ts = utility_functions.two_tree_ts()
-        algo, priors = self.run_inside_algorithm(ts, "gamma", standardize=False)
+        algo, priors, marg_lik = self.run_inside_algorithm(
+            ts, "gamma", standardize=False
+        )
         mut_rate = 0.5
         # priors[3][1] * Ll_(0->3)(1.2 - 0 + eps) ** 2
         node3_t1 = (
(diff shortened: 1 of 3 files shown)


nspope:
Ready for a look whenever -- any thoughts on this w.r.t. #237, @hyanwong?

As far as I can tell, using the marginal likelihood as a loss with which to tune the prior Ne seems to work pretty well. Here's an example with 1Mb/100 diploid simulations, where "RAE" stands for relative absolute error. The x-axis is the best possible RAE (e.g. pick the Ne that gives the lowest error against the true node ages), while the y-axis is the RAE obtained from tuning (e.g. pick the Ne that gives the highest marginal likelihood).



I haven't implemented the optimization itself in this PR-- but, this could potentially be done in `core.get_dates`, toggled via an optional flag. However, it's not clear how optimization would work in the context of an arbitrary population size history. My inclination would be to only optimize a single scaling term, analogous to the "ancestral population size" in most demographic inference methods, which should be enough to improve dating accuracy substantially without complicating things overmuch. So, the user would be supplying a "relative Ne trajectory" (or perhaps this could be calculated from pairwise coalescence rates).

nspope:
After rebasing this on #248, I've started to put together an iterative way to jointly infer dates and population size trajectories, using the coalescence rate machinery in [https://github.com/tskit-dev/tskit/pull/2621](https://github.com/tskit-dev/tskit/pull/2621). It works surprisingly well for a first pass.

The strategy is to initialize with a constant demography, where Ne is estimated from site diversity. Then use inside/outside algo to get dates. Then use `ts.coalescence_time_distribution` to get coalescence rates in epochs, where the epoch breaks are chosen to have roughly the same number of pairwise coalescence events (e.g. equal quantiles). The marginal likelihood is tracked at each iteration, and the algorithm terminates as soon as it starts to decrease. Optionally, I rescale the prior timegrid by optimizing marginal likelihood (that's what "rescaling" refers to in plots below).

As a test case, here's a 10Mb/200haps simulation from the *Bos taurus* model in `stdpopsim`, as this demography contains a very extreme recent decline (ping @gregorgorjanc). I compared four choices of prior:

1. "constant Ne" -- use a single epoch with Ne estimated from site diversity as the prior
2. "iter w/ rescaling" -- the iterative procedure described above with timegrid rescaling in each iteration
3. "iter w/o rescaling" -- the iterative procedure described above without the rescaling
4. "true demography" -- the true model is used as a prior

Here's relative absolute error against the true node times (left plot) and marginal likelihood (right plot). You can see that after a single iteration, the error drops substantially, and this is mirrored in the marginal likelihood (e.g. the latter is a reasonable proxy for the former, for a stopping rule).



Here are the estimated node ages:



Finally, I used the dated tree sequences to (re)infer the demography, again using coalescence rates (the true demography is the black line in each panel; colored lines are the trajectories inferred from dated tree sequences using one of the four priors listed above):



In this particular example, the rescaling didn't appear to help much (and increased the computational cost by an order of magnitude), but it does seem to help a lot in some cases (and this is reflected in the marginal likelihood -- in this case, the two iterative approaches produce very similar goodness-of-fit). It's also heartening that the constant Ne case is not totally off -- using site diversity to calibrate the prior really does seem to be a good default strategy.

Because this all depends on another PR that's not yet merged into tskit, I have to do a bit of work before integrating this into `date`.

gregorgorjanc:
@nspope whoa! This looks very good!!!! Even for the challenging BosTau case with the huge rapid recent decline in Ne! Just to add that in the last case where you used dated tree sequence, while all cases seem to be doing quite well, even the constant Ne, the other three do better for the rapid recent decline in Ne! That part is always challenging for methods like PSMC, which is one of the reasons method like GONE (LD based) is quite popular in our field (since Ne change can be drastic). Can we help with testing?

gregorgorjanc:
@nspope just to be clear - in this last case you take tree sequence that is dated from the simulation or you dated it with tsdate?

nspope:
@gregorgorjanc, yup all of those examples are dated with tsdate, just using different priors. With the true tree sequence (the true node ages) the inferred demography would be more or less exact, except for the terminal bins where there's always some artefacts.

The tree sequence topology is exact, though, not inferred with tsinfer -- so that's bound to introduce some error. I'm going to tinker with this strategy some more before it's ready to evaluate in earnest, but I'll keep you posted!

hyanwong:
This LGTM (I'm happy to change the return signature of `get_dates`, given that it isn't publicly documented). However, I assume that there was a reason to "normalise" (now called "standardise") the likelihoods by default (i.e. keeping `standardize/normalise=True` when creating a Likelihood object). With some of the weird inferred tree sequence topologies that @awohns was encountering with the unified genealogy, we ended up having to "normalise" stuff all over the place.

Perhaps @awohns could comment? Maybe a good thing to do would be to re-run tsdate with your PR on (say) the chromosomes in the unified genealogy, to check that we don't hit overflows. In fact, we wanted to do this anyway, to generate a new set of unified genealogies under the new `tsinfer` root-breaking idea.

awohns:
So if I recall correctly (we had a lot of iterations on this, so it's possible I'm wrong here), we didn't actually end up needing to normalise to prevent overflow after correctly implementing the log-sum-exp...
As you say @hyanwong the torture test here is the real inferred genealogies, as they will occasionally have pathological topologies, especially the "garbage collector" grand MRCA

hyanwong:
Thanks for the fast reply @awohns, and true about the grand MRCA. So I guess I could try re-dating (say) a small and a large chromosome under both this PR and the older `tsdate`, and if both pass (and give the same answers, to within a reasonable tolerance), we're happy to merge?

nspope:
Sounds good. The alternative is to store the normalising factor in the `Likelihood` class and combine it with the marginal likelihood later.

hyanwong:
Testing on the unified genealogies, it seems like standardisation is not needed, so I think we should resolve and merge this.

hyanwong:
Happy to merge this: could you possibly resolve the conflict @nspope?

nspope:
Oh, sure-- I'll just have it return nan for the variational algorithm until I figure out what the marginal likelihood is in that case.

nspope:
I think this is good to go!

PR #257, opened 2023-03-23
Gamma-variational inference scheme

nspope:
WIP

- Doesn't use a time discretization -- instead, each node has two variational parameters (shape/rate of gamma) for each of inside/outside/posterior
- The core idea is to approximate convolution likelihoods by "quasi-Poisson-like" functions. Then the updates are just summations of sufficient statistics for each incoming edge at each node.
- There's no numerical convolution required -- instead, everything is done with Laplace transforms (this requires evaluating hypergeometric functions, added `mpmath` as a dependency to do this as it seems to be much more stable numerically than `scipy`'s implementations)
- Some incoming features will be trickier to implement than for the discretized case (ancient samples, variable population size), but will still be possible I think

### DIFF
--- a/tests/distribution_functions.py
+++ b/tests/distribution_functions.py
@@ -1,0 +1,541 @@
+# MIT License
+#
+# Copyright (C) 2023 University of Oxford
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Utility functions to construct distributions used in variational inference,
+for testing purposes
+"""
+import mpmath
+import numpy as np
+import scipy.integrate
+import scipy.special
+
+from tsdate import approx
+from tsdate import hypergeo
+
+
+def kl_divergence(p, logq):
+    """
+    KL(p||q) divergence for fixed p
+    """
+    val, _ = scipy.integrate.quad(lambda x: -p(x) * logq(x), 0, np.inf)
+    return val
+
+
+def conditional_coalescent_pdf(t, n, k):
+    """
+    Evaluate the PDF of the conditional coalescent, using results from Wiuf and
+    Donnelly (1999).
+    """
+
+    def pr_t_bar_a(t, a, n):
+        """
+        Phase-type distribution for sum of waiting times during which
+        there are n, n-1, ..., a+1 ancestors.
+        """
+        lineages = np.arange(n, a, -1)
+        rate = (lineages * (lineages - 1)) / 2
+        weight = []
+        for i in range(len(rate)):
+            weight.append(1)
+            for j in range(len(rate)):
+                if i != j:
+                    weight[-1] *= rate[j] / (rate[j] - rate[i])
+        weight = np.array(weight)
+        val = np.sum(weight * rate * np.exp(-rate * t))
+        return val
+
+    def pr_a(a, n, k):
+        """
+        Hypergeometric-ish distribution for number of ancestors "a" after
+        subsample of size "k" has coalesced in tree of size "n"
+        """
+        return np.exp(
+            np.log(n + 1)
+            + scipy.special.betaln(n - k, k + 2)
+            - np.log(n - a)
+            - scipy.special.betaln(n - a - k + 2, k - 1)
+            - np.log(a + 1)
+            - scipy.special.betaln(a - 1, 3)
+        )
+
+    if n == k:
+        return pr_t_bar_a(t, 1)
+    else:
+        return np.sum(
+            [pr_a(a, n, k) * pr_t_bar_a(t, a, n) for a in range(2, n - k + 2)]
+        )
+
+
+class TiltedGammaDiff:
+    r"""
+    The distribution of the difference of two gamma RVs restricted to the
+    positive reals, tilted by an exponential function.
+
+    Specifically, for p(x; a, b) = b^a/Gamma(a) x^{a-1} e^{-xb}, this is
+
+        p(u; a1, a2, a3, b1, b2, b3) \propto p(u, a3, b3) \times
+            \int_u^inf p(x; a1, b1) p(x - u; a2, b2) dx, u \geq 0
+    """
+
+    @staticmethod
+    def _2F1(a, b, c, z):
+        """
+        log of |Re| for Gaussian hypergeometric function
+        """
+        val = mpmath.log(mpmath.hyp2f1(a, b, c, z))
+        return float(val)
+
+    @staticmethod
+    def _U(a, b, z):
+        """
+        log of |Re| for confluent hypergeometric function of the second kind
+        """
+        val = mpmath.log(mpmath.hyperu(a, b, z))
+        return float(val)
+
+    def __init__(self, shape1, shape2, shape3, rate1, rate2, rate3, reorder=True):
+        assert shape1 > 0 and shape2 > 0 and shape3 > 0
+        assert rate1 >= 0 and rate2 > 0 and rate3 >= 0
+        # for convergence of 2F1, we need rate2 > rate3. Invariant
+        # transformations of 2F1 allow us to switch arguments, with
+        # appropriate rescaling
+        self.reparametrize = rate3 > rate2
+        self.shape1 = shape1
+        self.shape2 = shape2
+        self.shape3 = shape3
+        self.rate1 = rate1
+        self.rate2 = rate2
+        self.rate3 = rate3
+        mu = (1.0 - shape1 - shape2) / 2
+        nu = (shape3 - 1) + (shape1 + shape2) / 2
+        zeta = 1 / 2 - (rate2 - rate3) / (rate1 + rate2)
+        assert zeta > -1 / 2, "Invalid parameters"
+        assert nu + 1 / 2 > abs(mu), "Invalid parameters"
+
+    def pdf(self, x):
+        r"""
+        Probability density function, derived using equation 3.383.4 from
+        Gradshteyn & Ryzhik (2000) "Table of Integrals, Series, and Products"
+
+        That is:
+
+        u^{a3 - 1} e^{(b2 - b3) u} \times
+          \int_u^inf x^{a1-1} (x-u)^{a2-1} e^{-x(b1+b2)} dx =
+        u^{a3 - 1} e^{(b2 - b3) u} (b1+b2)^{-(a1+a2)/2} Gamma(a2) \times
+          u^{(a1+a2)/2-1} e^{-(b1+b2)/2 u} W((a1-a2)/2, (1-a1-a2)/2, (b1+b2)u) =
+        (b1+b2)^{-(a1+a2)/2} Gamma(a2) u^{a3 + (a1+a2)/2 - 2} \times
+          e^{-u(b1+b2)(1/2+(b3 - b2)/(b1 + b2))} \times
+          W((a1-a2)/2, (1-a1-a2)/2, (b1+b2)u)
+
+        s.t. a2 > 0, (b1+b2)u > 0
+
+        Change variables, t = (b1+b2)*u and integrate t using 13.23.4 of DLMF,
+        to find the normalizing constant:
+
+        (b1+b2)^{-(a1+a2)/2} Gamma(a2) (b1+b2)^{-a3-(a1+a2)/2+2} \times
+          \int_0^inf t^{a3+(a1+a2)/2-2} e^{-t (1/2+(b3-b2)/(b1+b2))} \times
+          W((a1-a2)/2, (1-a1-a2)/2, t) 1/(b1+b2) dt =
+        (b1+b2)^{-a1-a2-a3+1} Gamma(a1+a2+a3-1) Beta(a2,a3) \times
+            2F1(a1+a2+a3-1, a3, a3+a2, (b2-b3)/(b1+b2))
+
+        Note that 2F1 is regularized in this equation, e.g. 2F1(a,b;c;z) =
+        2F1(a,b;c;z)/Gamma(c)
+
+        Transform W to U, and use Kummer's transformations:
+
+          W((a1-a2)/2, (1-a1-a2)/2, (b1+b2)u) =
+          exp(-u(b1+b2)/2) (b1+b2)^(1-a1/2-a2/2) \times
+            u^(1-a1/2-a2/2) U(1-a1, 2-a1-a2, (b1+b2)u) =
+          exp(-u(b1+b2)/2) (b1+b2)^(a1/2+a2/2) \times
+            u^(a1/2+a2/2) U(a2, a1+a2, (b1+b2)u)
+
+        so that the kernel becomes,
+
+          Gamma(a2) u^(a3 + a1 + a2 - 2) \times
+            e^(-u(b1+b3)) U(a2, a1+a2, (b1+b2)u)
+
+        In the special case where b1 == b3 == 0 and a1 == a3 == 1, the integral
+        becomes (via 3.382.2 GR2000):
+
+            e^{u b2} \int_u^inf (x-u)^{a2-1} e^{-x b2} dx = Gamma(a2) b2^{-a2}
+
+        e.g. a constant (this should never happen if the edge update order is
+        correct and samples are fixed to the present day, or if a proper prior
+        is used)
+        """
+        assert x >= 0.0, "PDF is only defined for non-negative reals"
+        A = self.shape1 + self.shape2 + self.shape3 - 1
+        B = self.shape3
+        C = self.shape2 + self.shape3
+        S = self.rate2 - self.rate3
+        T = self.rate2 + self.rate1
+        if self.reparametrize:
+            val = (
+                scipy.stats.gamma.logpdf(x, A, scale=1 / (self.rate1 + self.rate3))
+                + self._U(
+                    self.shape2,
+                    self.shape1 + self.shape2,
+                    (self.rate1 + self.rate2) * x,
+                )
+                + scipy.special.loggamma(self.shape2 + self.shape3)
+                - scipy.special.loggamma(self.shape3)
+                - self._2F1(A, C - B, C, S / (S - T))
+            )
+        else:
+            val = (
+                scipy.stats.gamma.logpdf(x, A, scale=1 / (self.rate1 + self.rate3))
+                + self._U(
+                    self.shape2,
+                    self.shape1 + self.shape2,
+                    (self.rate1 + self.rate2) * x,
+                )
+                + scipy.special.loggamma(self.shape2 + self.shape3)
+                - scipy.special.loggamma(self.shape3)
+                + A * np.log(T / (T - S))
+                - self._2F1(A, B, C, S / T)
+            )
+        return np.exp(val)
+
+    def laplace(self, s):
+        """
+        Laplace transform, derived using 13.23.4 in DLMF
+        """
+        # TODO: assert ROC
+        A = self.shape1 + self.shape2 + self.shape3 - 1
+        B = self.shape3
+        C = self.shape2 + self.shape3
+        S = self.rate2 - self.rate3
+        T = self.rate1 + self.rate2
+        if self.reparametrize:
+            val = (
+                -A * np.log(T - S + s)
+                + A * np.log(T - S)
+                + self._2F1(A, C - B, C, (s - S) / (T - S + s))
+                + -self._2F1(A, C - B, C, S / (S - T))
+            )
+        else:
+            val = self._2F1(A, B, C, (S - s) / T) - self._2F1(A, B, C, S / T)
+        return np.exp(val)
+
+    def moments(self):
+        r"""
+        Returns the first two raw moments.
+
+        Derived from differentiating the Laplace transform, and using the relation
+
+            d^n 2F1(A, B, C, z) / dz^n = a_n b_n / c_n 2F1(A + 1, B + 1, C + 1, z)
+
+        where x_n = x (x + 1) \dots (x + n - 1) is the rising factorial and 2F1 is the
+        Gaussian hypergeometric function. See 15.5.2 in DLMF.
+        """
+        A = self.shape1 + self.shape2 + self.shape3 - 1
+        B = self.shape3
+        C = self.shape2 + self.shape3
+        S = self.rate2 - self.rate3
+        T = self.rate1 + self.rate2
+        numer = 0
+        denom = 0
+        moments = []
+        if self.reparametrize:
+            F0 = self._2F1(A, C - B, C, S / (S - T)) - A * np.log(1 - S / T)
+            for i in range(2):
+                numer += np.log(A + i) + np.log(B + i) - np.log(C + i) - np.log(T)
+                denom = (
+                    F0
+                    - self._2F1(A + i + 1, C - B, C + i + 1, S / (S - T))
+                    + (A + i + 1) * np.log(1 - S / T)
+                )
+                moments.append(np.exp(numer - denom))
+        else:
+            F0 = self._2F1(A, B, C, S / T)
+            for i in range(2):
+                numer += np.log(A + i) + np.log(B + i) - np.log(C + i) - np.log(T)
+                denom = F0 - self._2F1(A + i + 1, B + i + 1, C + i + 1, S / T)
+                moments.append(np.exp(numer - denom))
+        return moments[0], moments[1]
+
+    def sufficient_statistics(self):
+        r"""
+        Return E[x], E[x^2], and E[log x].
+
+        To get raw moments use the Laplace transform E[e^{-su}]:
+
+        (b1+b2)^{-a1-a2-a3+1} Gamma(a1+a2+a3-1) Beta(a2,a3) \times
+            2F1(a1+a2+a3-1, a3, a3+a2, (b2-b3-s)/(b1+b2))
+
+        To get log moments use the Mellin transform E[u^s]:
+
+        (b1+b2)^{-a1-a2-a3-s+1} Gamma(a1+a2+a3+s-1) Beta(a2,a3+s) \times
+            2F1(a1+a2+a3+s-1, a3+s, a3+s+a2, (b2-b3)/(b1+b2))
+
+        In both cases: differentiate wrt s, evaluate at zero, and normalize
+        """
+        A = self.shape1 + self.shape2 + self.shape3 - 1
+        B = self.shape3
+        C = self.shape2 + self.shape3
+        S = self.rate2 - self.rate3
+        T = self.rate1 + self.rate2
+        if self.reparametrize:
+            # underflow protection is used, so the actual derivatives are
+            # np.exp(F) * dF_dz, etc.
+            F, dF_da, _, dF_dc, dF_dz, d2F_dz2 = hypergeo._hyp2f1_series(
+                A,
+                C - B,
+                C,
+                S / (S - T),
+            )
+            logconst = (
+                F
+                + scipy.special.betaln(self.shape2, self.shape3)
+                + scipy.special.loggamma(A)
+                - A * np.log(T - S)
+            )
+            x = -dF_dz * T / (S - T) ** 2 - A / (S - T)
+            xsq = (
+                d2F_dz2 * T**2 / (S - T) ** 4
+                + A * (A + 1) / (T - S) ** 2
+                + 2 * dF_dz * (1 + A) * T / (S - T) ** (3)
+            )
+            logx = (
+                (dF_da + dF_dc)
+                + -np.log(1 - S / T)
+                + scipy.special.digamma(A)
+                + scipy.special.digamma(B)
+                - scipy.special.digamma(C)
+                - np.log(T)
+            )
+        else:
+            F, dF_da, dF_db, dF_dc, dF_dz, d2F_dz2 = hypergeo._hyp2f1_series(
+                A,
+                B,
+                C,
+                S / T,
+            )
+            logconst = (
+                F
+                + scipy.special.betaln(self.shape2, self.shape3)
+                + scipy.special.loggamma(A)
+                - A * np.log(T)
+            )
+            x = dF_dz / T
+            xsq = d2F_dz2 / T**2
+            logx = (
+                (dF_da + dF_db + dF_dc)
+                + scipy.special.digamma(A)
+                + scipy.special.digamma(B)
+                - scipy.special.digamma(C)
+                - np.log(T)
+            )
+        return logconst, x, xsq, logx
+
+    def to_gamma(self, minimize_kl=True):
+        """
+        Return the shape and rate parameters of a gamma distribution with the
+        same expected sufficient statistics (if ``minimize_kl`` is ``True``), and
+        otherwise one with the same mean and variance.
+        """
+        logconst, x, xsq, logx = self.sufficient_statistics()
+        if minimize_kl:
+            _, xlogx, _ = approx.approximate_log_moments(x, xsq)
+            alpha, beta = approx.approximate_gamma_kl(x, logx, xlogx)
+        else:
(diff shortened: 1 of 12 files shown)


nspope:
I only have this working for a single tree (haven't quite figured out geometric scaling for this scheme yet), but it looks promising. Here's an example of 1000-haploid trees with increasing numbers of mutations:

<p float="left">
  
   
  
</p>


code to reproduce:
```python
import tsdate
import numpy as np
import msprime
import matplotlib.pyplot as plt

mu = 1e-8                                                                                                                                                                                      
ts = msprime.sim_ancestry(samples=500, population_size=10000, sequence_length=100000, recombination_rate=0)
ts = msprime.sim_mutations(ts, rate=mu)

foo = tsdate.date(ts, mutation_rate=mu, population_size=10000)
bar = tsdate.date(ts, mutation_rate=mu, population_size=10000, variational=True)

vanilla = foo.tables.nodes.time[ts.num_samples:]
variational = bar.tables.nodes.time[ts.num_samples:]
true = ts.tables.nodes.time[ts.num_samples:]

plt.scatter(np.log10(true), np.log10(vanilla), color='blue', label='tsdate')
plt.scatter(np.log10(true), np.log10(variational), color='red', label='tsdate-variational')
plt.axline((0, 0), slope=1, color="black", linestyle=(0, (5, 5)))
plt.legend()
plt.title(f"1000 haploids, 1 tree, {ts.num_mutations} mutations")
plt.xlabel("True node ages (log10)")
plt.ylabel("Inferred node ages (log10)")
plt.savefig(f"comp_{ts.num_mutations}.png")
plt.clf()
```

hyanwong:
This is is absolutely stellar work @nspope, thanks. I had got stuck on how to do the numerical convolution efficiently, so I'm really chuffed that you have an analytical solution. Let me know if there's anything I can do to help.

nspope:
Thanks @hyanwong -- it'd be nice to get some feedback on the design. I'm repurposing the `NodeGridValues` class to hold gamma parameters, and it'd probably be better to make a new class specifically for this task. Otherwise, does the organization seem clear?

nspope:
Now working with entire tree sequences -- it seems to do quite a bit better for the youngest and oldest nodes, and no worse otherwise:



Honestly I'm a bit shocked at how well this simple approximation scheme seems to work! 

As far as I'm concerned the algorithms are all there; what's left to do is write tests and tidy up code. Then, I can add support for variable population sizes in another PR, and (probably) speed up the per-node computations.

I think ancient samples will have to be supported via prior only -- that is, one can specify a gamma prior on the age of the sample. This could be made as precise as need be, but the limiting case where the prior becomes a delta function is not possible within this framework, I'm afraid.

hyanwong:
> Honestly I'm a bit shocked at how well this simple approximation scheme seems to work!

This is brilliant!
 
> As far as I'm concerned the algorithms are all there; what's left to do is write tests and tidy up code. 

Yes, I agree. I will look through this tonight and give feedback.

> Then, I can add support for variable population sizes in another PR, and (probably) speed up the per-node computations.

SGTM.

> I think ancient samples will have to be supported via prior only -- that is, one can specify a gamma prior on the age of the sample. This could be made as precise as need be, but the limiting case where the prior becomes a delta function is not possible within this framework, I'm afraid.

Quite a lot more complicated, but I wonder if we can support arbitrary functions for the prior, then use a quadrature scheme in their parent(s) to combine them with the arbitrary functions or gammas from their siblings, to make a gamma to allocate for the parents. In other words, we allow sample nodes (which are already treated as special) to have arbitrary or discretised distributions, but all non-sample nodes are approximated into a gamma.

Perhaps the proof of the pudding will be in the testing. It could easily be that the added complexity is not worth it.

nspope:
> I wonder if we can support arbitrary functions for the prior, then use a quadrature scheme in their parent(s) to combine them with the arbitrary functions or gammas from their siblings, to make a gamma to allocate for the parents. In other words, we allow sample nodes (which are already treated as special) to have arbitrary or discretised distributions, but all non-sample nodes are approximated into a gamma, so that the rest of the algorithm works as you have it now.

Yes, possibly. The constraint is that the support of the prior (whether for samples or internal nodes) needs to be over the positive reals-- because it ultimately needs to be approximated by a gamma.

hyanwong:
By the way, I assume that this would be (very?) hard to combine with more hidden states in the HMM. For example, if we simultaneously tried to estimate discrete locations on a lat/long grid, or (perhaps even better), a mean & variance of a normal distribution for latitude & a mean and variance for a normal distribution for longitude.

If we could solve this in any way, it would be a Great Leap Forward, IMO.

tsdate/approx.py
```diff
@@ -0,0 +1,325 @@
...
+        self.rate2 = rate2
+        self.beta = rate1 + rate2
+        self.mu = (1.0 - shape1 - shape2) / 2
+        self.kappa = (shape1 - shape2) / 2
+        self.nu = (shape1 + shape2) / 2
+        self.zeta = 1 / 2 - rate2 / (rate1 + rate2)
+        assert self.zeta > -1 / 2, "Invalid parameters"
+        assert self.nu + 1 / 2 > self.mu, "Invalid parameters"
+
+    def simulate(self, n):
```
hyanwong, line 113: Should this go in a separate test suite? If not, perhaps a comment to say that this is for testing only, I presume?

nspope:
Well ... gamma is conjugate prior for precision in a Gaussian, and presumably (if dispersal process is Brownian) then time would be something like the inverse of precision in the 2d Gaussian for locations. So maybe it'd be tractable. We should talk it over after benchmarking this a bit more.

nspope:
> Should we refactor somehow to abstract the logic into a single method that uses different passed-in classes?

Hrm, yes good idea -- I think we can pull the details of how the updates are done (that is, the convolutions) into the likelihood classes. Shall I do that in this PR or a followup (to make it easier to review)?

hyanwong:
> > Should we refactor somehow to abstract the logic into a single method that uses different passed-in classes?
> 
> Hrm, yes good idea -- I think we can pull the details of how the updates are done (that is, the convolutions) into the likelihood classes. Shall I do that in this PR or a followup (to make it easier to review)?

I think in this PR is fine.

nspope:
>   >  Shall I do that in this PR or a followup (to make it easier to review)?

On second thought, I'm going the keep the variational inside/outside code blocks separated out for now -- I think a see a way to make this an iterative, expectation-propagation scheme, and it'd be nice to be able to tinker without worrying about breaking the discretized algorithm.

hyanwong:
> > > Shall I do that in this PR or a followup (to make it easier to review)?
> 
> On second thought, I'm going the keep the variational inside/outside code blocks separated out for now -- I think a see a way to make this an iterative, expectation-propagation scheme, and it'd be nice to be able to tinker without worrying about breaking the discretized algorithm.

Yes, that sounds fine to me. We can revisit code abstraction when you have worked out the general principles.

nspope:
This is ready for another look @hyanwong -- the numerics have been a pain, with packaged special functions there's a tradeoff between using a very slow, high-precision ${_2}F_1$ that always works; and much faster implementations (like scipy's) that periodically return the wrong answer without warning. I think I've nailed down something that seems reasonably fast (3-4x better than original algo) and is numerically stable.

Once this is merged, I'd like to mess with the prior -- here's the comparison between the original algorithm, expectation propagation with the default prior, and EP with a "calibrated" global prior:

foo

Given that the latter case does substantially better whilst using the same prior for every node, I think there's room for improvement.

hyanwong:
> This is ready for another look @hyanwong ... 

Great, thanks for persisting with this.

> The numerics have been a pain... I think I've nailed down something that seems reasonably fast (3-4x better than original algo) and is numerically stable.

That sounds very tedious, and tricky both to diagnose and to test. Are there test cases in the PR to catch instabilities? I assume that is actually very difficult.

> Once this is merged, I'd like to mess with the prior 

Yes, I think that's a good plan. I agree that it seems like there is considerable scope for improvement there, and it's probably the major source of bias / noise at the moment.

hyanwong:
P.s. a few tiny linting bugs:

```
tsdate/approx.py:159:5: E722 do not use bare 'except'
tsdate/core.py:1006:5: E265 block comment should start with '# '
```

nspope:
> That sounds very tedious, and tricky both to diagnose and to test. Are there test cases in the PR to catch instabilities? I assume that is actually very difficult.

There are tests, but the cause of numeric instability is actually pretty simple to diagnose -- ${_2}F_1(a, b; c; z)$ (which shows up in the Laplace transform used in this variational method) is typically evaluated through it's series expansion. If any of the parameters are negative, then the series can have alternating terms; and if these terms are large then there's issues with floating point cancellation that are hard to fix without arbitrary precision arithmetic (which is much much slower). However, for our use case, the function can always be transformed into one with positive parameters -- although these non-alternating series will often converge more slowly, they are numerically stable and still much faster than multiprecision arithmetic (at least, without bringing in a heavy external C or Fortran dependency). I'm sure there's still ways to improve performance but this is a good starting point.

Er, long story short, I think I've got a handle on the problem and it should be usable after some code cleanup.

nspope:
This is ready to go, @hyanwong

nspope:
While looking into convergence criteria for this EP algorithm, I've been trying to assess how well gamma is doing, as a choice of posterior variational family. To do this, I take a grid of target coverage values from 0 to 1, call a given coverage $p$; calculate $p$-level "confidence intervals" for each node from quantiles of the marginal posterior; and calculate the proportion of nodes where the true age is within the interval.

The idea is that if the variational distribution is a good match overall, then the empirical coverage (e.g. the proportion of nodes with true ages covered by the $p$-level CI) should match the expected coverage $p$. So, we get something like a quantile-quantile plot. The animation below shows how the variational (marginal) posteriors improve over iterations:




There's definitely some deviation in the middle of the distribution, and the coverage is a little narrower than the ideal, but overall it seems like a really good fit. In other words, I think the variational posteriors give a pretty accurate picture of uncertainty in node ages.

hyanwong:
That's great news and a really good way to think about the fit. Probably worth thinking about how to get it into a publication (perhaps a single graph with shades of black to indicate number of iterations)

tsdate/core.py
```diff
@@ -523,6 +524,78 @@ def scale_geometric(self, fraction, value):
         return fraction * value
 
 
+class VariationalLikelihoods:
+    """
+    A class to store and process likelihoods for use in variational inference.
```
hyanwong, line 529: I think this stores the gamma in Laplace space, right? So that we can do the convolutions by addition. Should we say that up top?

tsdate/core.py
```diff
@@ -523,6 +524,78 @@ def scale_geometric(self, fraction, value):
...
+    def get_mut_edges(ts):
+        """
+        Get the number of mutations on each edge in the tree sequence.
+        """
+        return Likelihoods.get_mut_edges(ts)
+
+    @staticmethod
+    def combine(base, message):
+        """
+        Multiply two gamma PDFs
```
hyanwong, line 575: I'm not 100% sure how this works (i.e. where the [-1,0] comes from). We should probably state what format the "base" and "message" parameters are (i.e. a pair of numbers, representing the alpha and beta of the gamma distribution, I think?)

tsdate/core.py
```diff
@@ -858,6 +931,129 @@ def outside_maximization(self, *, eps, progress=None):
         return self.lik.timepoints[np.array(maximized_node_times).astype("int")]
 
 
+class ExpectationPropagation(InOutAlgorithms):
+    """
+    Expectation propagation, where the edge factors are approximated
```
hyanwong, line 936: What is meant by "edge factors" here?

tsdate/core.py
```diff
@@ -858,6 +931,129 @@ def outside_maximization(self, *, eps, progress=None):
...
+        """
+        Edges in order of increasing age of parent
+        """
+        return self.ts.edges()
+
+    def edges_by_child_desc(self):
+        """
+        Edges in order of decreasing age of child
+        """
+        wtype = np.dtype(
```
hyanwong, line 989: I found a nicer way to do this (https://github.com/tskit-dev/tskit/discussions/2744), simply something like:

```python
return (self.ts.edge(u) for u in np.lexsort((ts.edges_child, ts.nodes_time[ts.edges_child])))
```

tsdate/core.py
```diff
@@ -940,6 +1136,8 @@ def date(
     Ne=None,
     return_posteriors=None,
     progress=False,
+    expectation_propagation=0,  # TODO document
```
hyanwong, line 1139: This serves two roles, to turn on the EP routine and also as the number of iterations, I think? Is it also reasonable to be able to iterate using the old ("inside outside") algorithm? In which case we should probably have 2 parameters, firstly `num_iterations`, and secondly `method`?

I'm also slightly unclear how this related to the `outside maximisation` method that previously existed. Would it be worth working out a scheme for how they all fit together (we can do this later, as long as the current routines are not released, or are undocumented)

tsdate/core.py
```diff
@@ -858,6 +931,129 @@ def outside_maximization(self, *, eps, progress=None):
...
+        )
+        return sorted_child_parent
+
+    def propagate(self, *, edges, progress=None):
+        """
+        Update approximating factor for each edge
+        """
+        if progress is None:
+            progress = self.progress
+        # TODO: this will still converge if parallelized (potentially slower)
```
hyanwong, line 1010: Why potentially slower? Presumably because you would have to wait to update everything? @benjeffery has been doing some work over in tsinfer about parallelising work over independent edges, but my guess would be that in the case of tsdate, it's so fast as to not be worth the hassle working out the parallelisation scheme

tsdate/core.py
```diff
@@ -858,6 +931,129 @@ def outside_maximization(self, *, eps, progress=None):
...
+            total=self.ts.num_edges,
+            disable=not progress,
+        ):
+            if edge.child in self.fixednodes:
+                continue
+            if edge.parent in self.fixednodes:
+                raise ValueError("Internal nodes can not be fixed in EP algorithm")
+            edge_lik = self.lik.to_gamma(edge)
+            edge_lik += np.array([-1.0, 0.0])  # to Poisson, TODO cleanup
+            # cavity posteriors without approximate edge factor
```
hyanwong, line 1023: What is the cavity posterior. Is there an explanation somewhere?

tsdate/demography.py
```diff
@@ -144,6 +145,100 @@ def to_coalescent_timescale(self, time_ago):
...
+    #            gamma[0] * (t - n * a) ** 2
+    #            + 2 * gamma[1] * n * (t - n * a)
+    #            + gamma[2] * n**2
+    #        )
+    #    va -= mn**2
+    #    new_shape = mn**2 / va
+    #    new_rate = mn / va
+    #    return new_shape, new_rate
+
+    def to_gamma(self, shape=1, rate=1):
```
hyanwong, line 197: I don't think we use this in this PR do we? Should we put this into a follow up?

tests/distribution_functions.py
```diff
@@ -0,0 +1,541 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Utility functions to construct distributions used in variational inference,
```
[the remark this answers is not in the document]
nspope: Not quite -- what I've coded up here is the PDF of the conditional coalescence, whereas in priors.py there's only the moments. The PDF is needed to calculate KL divergence in the test suite.

The other two distributions in that file are not previously described, as far as I know. While they're exponential-family, they come about from a combination of tilting and convolution, and are written in terms of special functions (confluent and Gaussian hypergeometric functions). These classes actually aren't used for testing, but show the derivations for the distributions and various useful aspects of them (sufficient statistics, etc). So, we could remove these two at some point (for the moment, I'm referring to these implementations frequently).

tsdate/core.py
```diff
@@ -523,6 +524,78 @@ def scale_geometric(self, fraction, value):
         return fraction * value
 
 
+class VariationalLikelihoods:
+    """
+    A class to store and process likelihoods for use in variational inference.
+    """
+
+    probability_space = base.PAR
```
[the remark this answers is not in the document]
nspope: "parameter" as it's a grid of parameters rather than weights at time points. But happy to change it (or create a new class)

tsdate/core.py
nspope: Not in Laplace space-- "exponential family parameter" space (e.g. if you multiply / divide two gamma pdfs, that's the same as adding/subtracting their natural parameters). I can add some description

tsdate/core.py
nspope: The [-1, 0] is the conversion between the typical gamma parameterization and it's natural (exponential-family) parameterization. Can definitely add more description

tsdate/core.py
nspope: Poisson likelihoods for edges -- worth writing out some terminology here

tsdate/core.py
nspope: You mean, for a future version of tsdate where there's an interative algo for the old inside-outside method?

From the API side, maybe it's better to have a "control" argument that takes a dict of settings, that are used (or not) by whatever "method" is requested? Similar to what scipy does for its high level optimization functions. Then we won't have to translate every option across algorithms.

tsdate/core.py
nspope: I just mean the rate of convergence will be slower, because the way in which edges are currently traversed (serially by time) is optimal. When parallelized, it'll break this order.

The algorithm overall might be much faster when parallelized, it just might take more iterations.

tsdate/core.py
nspope: The idea is that you can remove the contribution of an edge to the global approximation by subtracting natural parameters. The resulting reduced approximation is called the "cavity" (for a particular edge) in the BP/EP literature. Then, moment matching is performed against the cavity + the exact edge likelihood, to get a new "approximate edge factor".

I can add some description to the docstrings.

tsdate/demography.py
nspope: It's used here https://github.com/tskit-dev/tsdate/blob/1fbfcba4965c0383ad6ea78742c5effc2576aade/tsdate/prior.py#L1134-L1136

and also has some tests written. It's necessary in fact even when there's no variation in population size, to get the scaling right

nspope:
Thanks for looking this over @GertjanBisschop! I think it'd help if we could meet and go over details, but I'll give a quick overview. It'll help if I briefly describe the algorithm and numerical bottleneck. 

Basically, we're trying to match expected sufficient statistics between the approximation (a product of gamma distributions, one for each node) to their expectations under the exact model (a product of gamma priors for each node and Poisson likelihoods per edge). This is impossible to do globally (indeed, if we could do it globally then we could do exact inference). Instead, EP replaces the global problem with a series of local problems, one for each edge. I won't lay out the derivation, but it boils down to finding the expectations of the sufficient statistics $\phi(t_i, t_j) = [t_i, \log t_i, t_j, \log t_j]$ where $i,j$ are parent and child of the edge, under the density $$[q^{\setminus ij} p_{ij}] (t_i, t_j) \propto \mathbb{I}[t_i > t_j] t_i^{\alpha_i - 1} e^{-\beta_i t_i} t_j^{\alpha_j - 1} e^{-\beta_j t_j} (t_i - t_j)^{y_{ij}} e^{-(t_i - t_j) \mu_{ij}}$$  Here $q^{\setminus ij}$ is called the "cavity distribution" and $p_{ij}$ is the exact edge likelihood, and the product of the two (the expression above) is the "target". $\alpha_i, \beta_i, \alpha_j, \beta_j$ are variational parameters, $y_{ij}$ is the edge mutation count, and $\mu_{ij}$ is the mutation rate on the edge. (Just noting in  passing that if you removed the $t_i^{\alpha_i - 1} e^{-\beta_i t_i}$ part and integrated over $t_j$ then you'd have the distribution of a sum of two gammas -- but, with that term in there, it's a more complicated and obscure distribution).

So, to rephrase, we need to get $\mathbb{E}[\phi(t_i, t_j)] = \int_0^\infty \int_0^\infty \phi(t_i, t_j) [q^{\setminus ij} p_{ij}] (t_i, t_j) dt_i dt_j$. The target $q^{\setminus ij} p_{ij}$ is not a standard density, but it is an exponential family (that is, it can be written as an expression that's log-linear in the parameters and a vector of statistics that include the sufficient statistics we're after). So, all we need to do is find the log-partition function, $A(\alpha_i, \beta_i, \alpha_j, \beta_j) = \log \int \int [q^{\setminus ij} p_{ij}] (t_i, t_j) dt_i dt_j$ and differentiate it with respect to $\alpha, \beta$ to get the required sufficient statistics. It can be proved (with a lot of tedious manipulation of special functions) that this has the form,

$$
\begin{aligned}
\exp A(\alpha_i, \beta_i, \alpha_j, \beta_j) = {\_2}F_1(\alpha_j, \alpha_i + \alpha_j + y_{ij}; \alpha_j + y_{ij}; (\mu_{ij} - \beta_j) (\mu_{ij} + \beta_i)^{-1}) \\
\times \mathcal{B}(y_{ij} + 1, \alpha_j) \Gamma(\alpha_i + \alpha_j + y_{ij}) (\mu_{ij} + b_i)^{-\alpha_i - \alpha_j - y_{ij}}
\end{aligned}
$$

The tricky part of this is differentiating ${_2}F_1$ with respect to its parameters (it's easy for the last parameter -- it's "argument" -- but not for the other parameters). There's also issues with overflow in canned implementations of ${_2}F_1$, like scipy's. It turns out that the fastest yet numerically stable option is to take its Taylor series and differentiate term by term until convergence. We can then check the answer is valid by plugging the derivatives into the differential equation the defines ${_2}F_1$. Along the way, we use equivalent transformations of ${_2}F_1$ to get it into a form where the Taylor series converges quickly. Most of the new code/test revolves around these numerical issues.

There's a lot that's different about this scheme from the stuff I presented back in tskit meeting, so it'd be good to meet and go over details. That'd probably help alot with making sense of the code.

tsdate/core.py
hyanwong: Good point about a 'control' argument. I like that.

tsdate/core.py
hyanwong: That would be great, thanks Nate.

tsdate/demography.py
hyanwong: Oh, sorry, I missed that.

tests/test_approximations.py
```diff
@@ -0,0 +1,188 @@
...
+        # compare KL divergence between strategies
+        kl_0 = kl_divergence(
+            lambda x: conditional_coalescent_pdf(x, self.n, k),
+            lambda x: scipy.stats.gamma.logpdf(x, alpha_0, scale=1 / beta_0),
+        )
+        kl_1 = kl_divergence(
+            lambda x: conditional_coalescent_pdf(x, self.n, k),
+            lambda x: scipy.stats.gamma.logpdf(x, alpha_1, scale=1 / beta_1),
+        )
+        print(kl_1, kl_0)
```
hyanwong, line 187: Stray print here

tests/test_approximations.py
```diff
@@ -0,0 +1,188 @@
...
+                t_i ** (a_i - 1)
+                * np.exp(-t_i * b_i)
+                * t_j ** (a_j - 1)
+                * np.exp(-t_j * b_j)
+                * (t_i - t_j) ** y
+                * np.exp(-(t_i - t_j) * mu)
+            )
+
+    def test_sufficient_statistics(self, pars):
+        print((pars[5] - pars[3]) / (pars[1] + pars[5]))
```
hyanwong, line 73: Stray print here

tests/test_hypergeo.py
```diff
@@ -0,0 +1,248 @@
...
+            step=1e-7,
+            richardson_terms=4,
+        )
+        return grad(p)
+
+    def test_2f1(self, pars):
+        f, s, *_ = hypergeo._hyp2f1_taylor_series(*pars)
+        check = self._2f1_validate(*pars)
+        assert s == mpmath.sign(check)
+        print(f, mpmath.log(mpmath.fabs(check)))
```
hyanwong, line 93: Stray print

tests/test_hypergeo.py
```diff
@@ -0,0 +1,248 @@
...
+        assert np.isclose(f, float(mpmath.log(mpmath.fabs(check))))
+
+    def test_2f1_grad(self, pars):
+        _, _, *grad = hypergeo._hyp2f1_taylor_series(*pars)
+        grad = grad[:-1]
+        offset = self._2f1_validate(*pars)
+        check = self._2f1_grad_validate(*pars, offset=offset)
+        print(grad, check)
```
hyanwong, line 101: Stray print

tests/test_hypergeo.py
```diff
@@ -0,0 +1,248 @@
...
+            richardson_terms=4,
+        )
+        return grad(p)
+
+    def test_2f1(self, pars):
+        pars = self._transform_pars(*pars)
+        f, s, *_ = hypergeo._hyp2f1_recurrence(*pars)
+        check = self._2f1_validate(*pars)
+        assert s == mpmath.sign(check)
+        print(f, mpmath.log(mpmath.fabs(check)))
```
hyanwong, line 145: Stray print

tests/test_hypergeo.py
```diff
@@ -0,0 +1,248 @@
...
+        assert np.isclose(f, float(mpmath.log(mpmath.fabs(check))))
+
+    def test_2f1_grad(self, pars):
+        pars = self._transform_pars(*pars)
+        _, _, *grad = hypergeo._hyp2f1_recurrence(*pars)
+        grad = grad[:-1]
+        offset = self._2f1_validate(*pars)
+        check = self._2f1_grad_validate(*pars, offset=offset)
+        check[1] = 0.0  # integer parameter has no gradient
+        print(grad, check)
```
hyanwong, line 155: Stray print

tests/test_hypergeo.py
```diff
@@ -0,0 +1,248 @@
...
+        print(f, check)
+        assert np.isclose(f, check)
+
+    def test_2f1_grad(self, muts, hyp2f1_func, pars):
+        pars[4] = muts
+        _, s, *grad = hyp2f1_func(*pars)
+        assert s > 0
+        offset = self._2f1_validate(*pars)
+        check = self._2f1_grad_validate(*pars, offset=offset)
+        print(grad, check)
```
hyanwong, line 247: Stray print

hyanwong:
One other point. I guess it would be good to add some tests of the entire method (i.e. calling `tsdate.date(method="approx_gamma_EP", control={"iterations": 20}` etc. Should we add those before merging too? Probably worth providing some "bad" values for method, some bad "control" parameters, and to check that the number of iterations is correctly performed? Oh, and it would presumably be useful to make sure the method and control parameters are saved in the provenance of the resulting tree sequence?

If you don't have the capacity to do this, I can have a go at it (although TBH I'd prefer to spend my time writing documentation and tutorial content for the new functionality).

tsdate/core.py
```diff
@@ -1180,3 +1390,140 @@ def get_dates(
...
+    # Default to not creating approximate priors unless ts has > 1000 samples
+    approx_priors = False
+    if tree_sequence.num_samples > 1000:
+        approx_priors = True
+
+    if priors is None:
+        if population_size is None:
+            raise ValueError(
+                "Must specify population size if priors are not already "
+                "built using tsdate.build_parameter_grid()"
```
GertjanBisschop, line 1484: should this be `tsdate.build_prior_grid()`?

tsdate/core.py
```diff
@@ -1180,3 +1390,140 @@ def get_dates(
...
+    Infer dates for the nodes in a tree sequence, returning an array of inferred dates
+    for nodes, plus other variables such as the posteriors object
+    etc. Parameters are identical to the date() method, which calls this method, then
+    injects the resulting date estimates into the tree sequence
+
+    :return: a tuple of ``(mn_post, posteriors, timepoints, eps, nodes_to_date)``.
+        If the "inside_outside" method is used, ``posteriors`` will contain the
+        posterior probabilities for each node in each time slice, else the returned
+        variable will be ``None``.
+    """
```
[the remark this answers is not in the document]
GertjanBisschop: Aha I get it. This is code that is copied from `get_dates()`. Same holds for the next remark. Can code duplication be avoided here?

GertjanBisschop (approved):
Awesome work!

tsdate/core.py
nspope: `build_prior_grid` is what the original algo uses to make a set of timepoints from prior quantiles. This is separate from `build_parameter_grid` which dumps the prior parameters instead of a set timepoints. Personally, I think keeping them as separate functions is cleaner -- but I don't have a strong opinion.

tsdate/core.py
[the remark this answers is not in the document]
nspope: I think once we have an iterative version of the original algorithm, it'll totally be possible to consolidate. ATM though I think it'd be easier to keep the variational code separate.

nspope:
Thanks for the feedback @hyanwong and @GertjanBisschop!

With regards to code consolidation (@GertjanBisschop), I think if we could keep the variational stuff separate for the moment (e.g. `variational_dates` separated from `get_dates`) that'd make life easier in the short term. Ultimately (once the variational stuff has settled in a bit) I think we can definitely consolidate, especially if we include an iterative form of the original BP algorithm.

tsdate/approx.py
```diff
@@ -0,0 +1,186 @@
...
+    assert mean > 0
+    assert variance > 0
+    logx = np.log(mean) - 0.5 * variance / mean**2
+    xlogx = mean * np.log(mean) + 0.5 * variance / mean
+    logx2 = np.log(mean) ** 2 + (1 - np.log(mean)) * variance / mean**2
+    return logx, xlogx, logx2
+
+
+@numba.njit("UniTuple(float64, 2)(float64, float64)")
+def approximate_gamma_mom(mean, variance):
```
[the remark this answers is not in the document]
nspope: Good catch, guess that's not being used

tsdate/core.py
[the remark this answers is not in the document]
hyanwong: Yeah, I agree with keeping separate until things settle down. We can open an issue about consolidation if we want

tsdate/core.py
hyanwong: I've merged this (in the core.py_ file into the main tsdate branch

nspope:
I think this is good to go. I'm punting some things down the road (using `control` dict instead of kwargs, filling out the bare-bones testing) until I'm finished with the prior calibration issues. I'll open some issues to track these and other to-dos.

hyanwong (approved):
Isn't `edges_by_child_desc` etc used in the inside/outside also too?

Note that I simplified it a few days ago via #260

tsdate/demography.py
```diff
@@ -144,6 +145,100 @@ def to_coalescent_timescale(self, time_ago):
         )
         return coalescent_time_ago
 
+    # @staticmethod
+    # def _Gamma(z, a=0, b=np.inf):
```
hyanwong, line 149: I'm assuming that there's a reason you have left this in, but commented out?

tests/test_inference.py
```diff
@@ -372,3 +372,41 @@ def test_simple_sim_multi_tree(self):
...
+    """
+
+    def test_simple_sim_1_tree(self):
+        ts = msprime.simulate(8, mutation_rate=5, random_seed=2)
+        tsdate.date(ts, mutation_rate=5, population_size=1, method="variational_gamma")
+
+    def test_simple_sim_multi_tree(self):
+        ts = msprime.simulate(8, mutation_rate=5, recombination_rate=5, random_seed=2)
+        tsdate.date(ts, mutation_rate=5, population_size=1, method="variational_gamma")
+
```
hyanwong, line 389: Need to add an example of passing a control dictionary.

hyanwong:
I'm merging this for now. We can fix any follow ups in further PRs.

Note, however, that CI does bring up a few test warnings. I don't think we need to worry about these, but they might be worth digging into just to check the tests are correct.

```
=============================== warnings summary ===============================
tests/test_approximations.py::TestPosteriorMomentMatching::test_sufficient_statistics[pars0]
tests/test_approximations.py::TestPosteriorMomentMatching::test_sufficient_statistics[pars1]
tests/test_approximations.py::TestPosteriorMomentMatching::test_sufficient_statistics[pars2]
tests/test_approximations.py::TestPosteriorMomentMatching::test_sufficient_statistics[pars3]
  /home/circleci/.local/lib/python3.7/site-packages/scipy/integrate/quadpack.py:880: IntegrationWarning: The integral is probably divergent, or slowly convergent.
    **opt)

tests/test_approximations.py::TestPosteriorMomentMatching::test_sufficient_statistics[pars0]
  /home/circleci/.local/lib/python3.7/site-packages/scipy/integrate/quadpack.py:880: IntegrationWarning: The maximum number of subdivisions (50) has been achieved.
    If increasing the limit yields no improvement it is advised to analyze 
    the integrand in order to determine the difficulties.  If the position of a 
    local difficulty can be determined (singularity, discontinuity) one will 
    probably gain from splitting up the interval and calling the integrator 
    on the subranges.  Perhaps a special-purpose integrator should be used.
    **opt)
```

nspope:
Great, thanks. Those warnings are from a double quadrature in the tests, to check the moment calculations-- not from any of the actual tsdate infrastructure. I think we can get rid of them by increasing the number of subdivisions used in the quadrature routine, or something similar. Thankfully they don't affect the precision of the result to a noticeable degree, so safe to ignore for now.

hyanwong:
Warnings fixed in #269

PR #273, opened 2023-06-15
Raise errors not warnings if we can't convert a probability space

hyanwong:
And test the probability space conversion more


hyanwong:
No brainer. Merging

PR #283, opened 2023-06-21
Save PopulationSizeHistory params

### LINKED ISSUE #274: Store PopulationSizeHistory as a json list
It would be nice to store the passed-in PopulationSizeHistory object in the provenance of a tsdated tree sequence. I don't think it is a very large object, right? It just consists of 2 arrays, neither of which is likely to be millions of numbers long. Any reason why not to store it?

hyanwong:
Fixes #274

### DIFF
--- a/tsdate/demography.py
+++ b/tsdate/demography.py
@@ -69,39 +69,48 @@ class PopulationSizeHistory:
 
     def __init__(self, population_size, time_breaks=None):
         """
-        :param np.ndarray population_size: A numpy array containing diploid
+        :param array_like population_size: An array containing diploid
             population sizes per epoch
-        :param np.ndarray time_breaks: A sorted numpy array containing time
+        :param array_like time_breaks: A sorted array containing time
             breaks that divide epochs, measured in units of generations in the
             past
         """
 
         if time_breaks is None:
-            time_breaks = np.array([], dtype=float)
+            time_breaks = []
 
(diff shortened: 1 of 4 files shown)


hyanwong:
Does this look right @nspope ?

Also pinging @benjeffery , because I have previously wondered about the general strategy for encoding simple objects into the `tskit` provenance information, and this is one possibility, although it does involve allowing multiple types for a parameter (i.e. either a simple object of the correct type or a dictionary of parameters used to instantiate an object of the correct type). The case I was previously stuck on was whether/how to encode a recombination rate map into tsinfer provenance information. That's potentially a much bigger array, though, so maybe not worth it. See https://github.com/tskit-dev/tsinfer/issues/332#issuecomment-867792691

tsdate/demography.py
```diff
@@ -111,6 +120,17 @@ def __init__(self, population_size, time_breaks=None):
         self.coalescent_breaks = coalescent_breaks
         self.coalescent_rate = coalescent_rate
 
+    def as_init_params(self):
```
nspope, line 123: Maybe `as_dict` for this method name?

nspope (approved):
LGTM!

tsdate/demography.py
hyanwong: True. I guess this entirely contains all the info we need, so `as_dict` is correct?

Issue #292, opened 2023-07-11
Use of, and alternatives to, to a "global" (i.e. node-agnostic) prior


hyanwong:
@nspope has found problems with the conditional coalescent mixture priors used by tsdate. Surprisingly, better results are gained by averaging the priors over all the nodes, and using this single distribution as a prior for all nodes (e.g. see https://github.com/tskit-dev/tsdate/pull/257#issuecomment-1532446409).

I have opened up this issue as a place to discuss what the best prior strategy is. We should be able to do better than the naive global approach, but I'm not sure how!

(update Summer 2024: rather than constructing a global prior from the coalescent, we now use a "flat" prior for all except the root nodes - although this is actually not truly flat, as internal nodes are constrained by the position of the roots, see [#425](https://github.com/tskit-dev/tsdate/issues/425#issuecomment-2253765190). This makes tsdate much more robust to weird demographies etc, and allows for historical samples. Although for simple demographies with all-contemporary samples, setting coalescent-based priors on node times might give more power, for the moment we get better and more robust results without making coalescent assumptions, so this issue is more of a "nice-to-have")

hyanwong:
Here's an interesting way to plot the expected node times in a 4-tip tree sequence, depending on the mixture of sample descendants. We can use a barycentric (i.e. triangular) plot to represent the ratios of the span over which a node has 2 : 3 : 4 descendants, and then show separate plots for different total node span lengths. Something like this for 100 replicate simulations of 100Mb simulated tree sequences:

![image](https://github.com/tskit-dev/tsdate/assets/4699014/799d9113-42cc-4740-a319-4508d88c2f63)

code:

```python
import msprime
import numpy as np
import tqdm
from matplotlib import pyplot as plt
import scipy

reps = 100

# Save into a massive matrix with 4 columns:
# 2span 3span 4span time
nodes = None
for ts in tqdm.tqdm(
    msprime.sim_ancestry(4, ploidy=1, population_size=1e4, sequence_length=1e8, recombination_rate=1e-8, num_replicates=reps)
):
    node_arr = np.zeros((ts.num_nodes - ts.num_samples, ts.num_samples))
    node_arr[:, 3] = ts.nodes_time[ts.num_samples:]
    for tree in ts.trees():
        for u in tree.nodes():
            node_arr[u-ts.num_samples][tree.num_samples(u) - 2] += tree.interval.span
    if nodes is None:
        nodes = node_arr
    else:
[...]

PR #294, opened 2023-07-12
Expose node spans not weights

hyanwong:
And implement (but don't use) a log-span weighting option for mixture priors. See https://github.com/tskit-dev/tsdate/issues/292#issuecomment-1633112244


hyanwong:
This doesn't seem any slower in my tests, and reporting the spans before normalisation is more informative anyway, so I'm inclined to merge it, without changing the default weighting behaviour for the time being.

PR #297, opened 2023-07-14
Switch to auto tqdm

### LINKED ISSUE #101: make tqdm bars notebook friendly
tqdm bars print repeatedly in notebooks, can we fix this?

hyanwong:
Hopefully gives a nicer output in notebooks. Fixes #101


hyanwong:
Merging since @benjeffery thinks it should be OK :smile:

Issue #203, opened 2022-03-04
Set node metadata schema


hyanwong:
At the moment we don't set a metadata schema for nodes returned by `tsdate`: we simply dump a binary json string in there, e.g.

```
>>> dated_ts.node(node_id)
Node(id=37, flags=0, time=33273.69140479153, population=-1, individual=-1, metadata=b'{"mn": 33273.69140479153, "vr": 12870629.731027713}')
```

We should probably attempt to set the "mn" and "vr" keys assuming the schema is valid, and simply omit them (perhaps with a warning) if it is not. Perhaps we should also store this as `{"tsdate_time":{"mn":XXX,"vr":YYY}}` so that it's clear what the metadata refers to?

If no schema exists, and the node metadata is entirely empty, we can probably set the nodes table schema to `tskit.MetadataSchema.permissive_json()`

I wonder if @benjeffery, the metadata king, has any thoughts on the best thing to do here.

hyanwong:
I just remembered https://github.com/tskit-dev/tsinfer/issues/416, in which the metadata schema is set for the node table for tree sequences produced by tsinfer.

I noted there that if, for efficiency, we have a tsinfer-specific struct schema for node metadata, we should probably check for it in tsdate and amend the schema to add the necessary fields that tsdate wants to set.

jeromekelleher:
Just to note here that we're supposed to be working with things that **arent'** tsinfer too, so we need to be very careful about assumptions we make about metadata.

benjeffery:
Thanks for flagging this up @hyanwong - can I clarify that the aim here is to add tsdate specific attributes to the possibly-preexisting metadata? If so this would seem to be a great use case to consider as we add the higher level metadata API. 
As I won't get to that till this round of C work in tskit is done, maybe we should write some strawman metadata updating code here as first stab at what the metadata API will eventually do, then replace it when that is ready.
Basically something like "If there is an updateable schema, either stuct or json then add in the new properties, if not assign the minimal schema for the tsdate properties".

hyanwong:
> @benjeffery: can I clarify that the aim here is to add tsdate specific attributes to the possibly-preexisting metadata?

Yes, exactly.

> @jeromekelleher: Just to note here that we're supposed to be working with things that **arent'** tsinfer too, so we need to be very careful about assumptions we make about metadata.

[...]

PR #303, opened 2023-07-23
Set metadata schemas

### LINKED ISSUE #302: "set_metadata" parameter
Linked to #203 and https://github.com/tskit-dev/tsdate/issues/237#issuecomment-1438714926. At the moment, `tsdate`, as well as changing node times, also changes node metadata to add (unconstrained) means and variances. It would be useful to have the ability to (a) not add to metadata and (b) specify whether or not node metadata should be stomped on, if it exists.

Since this is the only metdata that tsdate touches, I suggest that as a first step, we could have a parameter `set_metadata`, `save_metadata`, `store_metadata` or even `force_metadata` with the following values:

* `False`: do not touch any metadata columns in any tables
* `True`: always add metadata to the node column, changing the schema and overwriting any existing node metadata
* `None` (default): try to append the existing node metadata, overwriting the tsdate-specific fields if they exist. A minor wrinkle is if there is binary node metadata (i.e. without a schema) and the metadata column is not empty (see https://github.com/tskit-dev/tsdate/issues/203#issuecomment-1438679651). In this case we won't know how to adjust the node metadata properly, so we should either raise an error or emit a warning. I suggest we should raise a warning, but it would be a pain to have run tsdate all the way through only to raise an error at the end, so we could check before running the main tsdate routine, and raise the error early?

hyanwong:
Defaults to a "struct" type unless a schema already exists (or if there is a null schema that can be interpreted as JSON). If so, the schema is modified to allow the mean time and variance in time (`mn` & `vr`) to be saved as extra values.

To maintain backwards-compatibility, I have used the top-level keys "mn" and "vr" for the mean and variance in times (the description field in the schema should make it clear what these are).

Fixes #302

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -1838,7 +1838,7 @@ class TestNodeTimes:
 
     def test_fails_unconstrained(self):
         ts = utility_functions.two_tree_mutation_ts()
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="must be tsdated"):
             nodes_time_unconstrained(ts)
 
 
@@ -1852,9 +1852,14 @@ class TestSiteTimes:
         with pytest.raises(ValueError):
             tsdate.sites_time_from_ts(ts)
 
+    def test_undated(self):
+        ts = utility_functions.two_tree_mutation_ts()
+        with pytest.raises(ValueError, match="Try calling"):
+            tsdate.sites_time_from_ts(ts, unconstrained=True)
+
     def test_node_selection_param(self):
         ts = utility_functions.two_tree_mutation_ts()
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="node_selection parameter"):
             tsdate.sites_time_from_ts(ts, node_selection="sibling")
 
     def test_sites_time_insideoutside(self):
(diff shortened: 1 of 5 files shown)


hyanwong:
Also fixes #203

benjeffery (commented):
Looks good @hyanwong, we really should get the metadata updating methods into tskit proper, but I think they'll have a similar behaviour to what you've done here.
How much do we care about backwards compatibility here? The two character names suck, and with `struct` there is no cost to longer, more descriptive names.

hyanwong:
Thanks @benjeffery. I agree that the 2 character names are a bit crap. I'm not that concerned about backwards compatibility, TBH, but at the moment, tsinfer sets JSON metadata on nodes, and with this PR to tsdate, we don't change the codec. So long names may kill us, storage-wise, when running tsdate on the current tsinferred trees.

If we address https://github.com/tskit-dev/tsinfer/issues/416 then this ceases to be an issue, as tsinfer would output struct metadata for nodes too. I'm inclined to merge this PR (which doesn't really change behaviour much), but open another to say that we should switch to a more legible naming scheme when tsinfer switches to struct?

I assume that you think `struct` on node metadata in tsinfer is a good call?

hyanwong:
Obsoleted by #383 . We can revisit using struct metadata later.

PR #306, opened 2023-08-03
Faster prior variance calculation

nspope:
Use a recursive approach to calculating prior variances that is a couple of orders of magnitude faster:

so it's possible to get the exact variances for 100,000 tips in a few minutes (by comparison, it's infeasible to get exact variances for more than 2,000 tips or so with the original algorithm). 

Some indication that the two algorithms are equivalent, showing sqrt(variance) for 1280 tips:

Still trying to get my head around the interpolation scheme used for the approximation, and if this algorithm will be useful there (it recurses backwards from the maximum number of tips).

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -63,28 +63,14 @@ class TestBasicFunctions:
     Test for some of the basic functions used in tsdate
     """
 
-    def test_alpha_prob(self):
-        assert ConditionalCoalescentTimes.m_prob(2, 2, 3) == 1.0
-        assert ConditionalCoalescentTimes.m_prob(2, 2, 4) == 0.25
-
     def test_tau_expect(self):
         assert ConditionalCoalescentTimes.tau_expect(10, 10) == 1.8
         assert ConditionalCoalescentTimes.tau_expect(10, 100) == 0.09
         assert ConditionalCoalescentTimes.tau_expect(100, 100) == 1.98
         assert ConditionalCoalescentTimes.tau_expect(5, 10) == 0.4
 
-    def test_tau_squared_conditional(self):
-        assert ConditionalCoalescentTimes.tau_squared_conditional(
-            1, 10
-        ) == pytest.approx(4.3981418)
-        assert ConditionalCoalescentTimes.tau_squared_conditional(
-            100, 100
-        ) == pytest.approx(4.87890977e-18)
-
-    def test_tau_var(self):
-        assert ConditionalCoalescentTimes.tau_var(2, 2) == 1
-        assert ConditionalCoalescentTimes.tau_var(10, 20) == pytest.approx(0.0922995960)
-        assert ConditionalCoalescentTimes.tau_var(50, 50) == pytest.approx(1.15946186)
+    def test_tau_var_mrca(self):
+        assert np.isclose(ConditionalCoalescentTimes.tau_var_mrca(50), 1.15946186)
 
     def test_gamma_approx(self):
         assert gamma_approx(2, 1) == (4.0, 2.0)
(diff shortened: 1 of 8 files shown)


nspope:
Some interesting (but possibly useless) applications of this algorithm:

- Get logarithmic moments -- this would allow matching lognormal/gamma via sufficient statistics rather than mean/variance
- Adjust for variable population size at the level of the hypoexponential components (e.g. do the time rescaling *before* marginalizing over number of extant ancestors, and mixing over node spans) -- this should be more accurate? But maybe not enough to prioritize?

hyanwong:
Really neat. Thanks for this Nate.

nspope:
Darn, this is working but fried some unit tests.

nspope:
This is good to go. I moved the original, slow variance calculation into `utility_functions` and test vs fast implementation.

tests/test_functions.py
```diff
@@ -63,28 +63,14 @@ class TestBasicFunctions:
     Test for some of the basic functions used in tsdate
     """
 
-    def test_alpha_prob(self):
```
hyanwong, line 66: Is it worth keeping these tests in a separate `test_utilities` file?

tests/test_functions.py
```diff
@@ -63,28 +63,14 @@ class TestBasicFunctions:
...
-        assert ConditionalCoalescentTimes.m_prob(2, 2, 3) == 1.0
-        assert ConditionalCoalescentTimes.m_prob(2, 2, 4) == 0.25
-
     def test_tau_expect(self):
         assert ConditionalCoalescentTimes.tau_expect(10, 10) == 1.8
         assert ConditionalCoalescentTimes.tau_expect(10, 100) == 0.09
         assert ConditionalCoalescentTimes.tau_expect(100, 100) == 1.98
         assert ConditionalCoalescentTimes.tau_expect(5, 10) == 0.4
 
-    def test_tau_squared_conditional(self):
```
hyanwong, line 76: Same here - maybe worth keeping to test the basic version

hyanwong (approved):
LGTM. Maybe it's worth keeping a few tests for the simpler version? Also, using the approximation only when N>20,000 seems quite a high threshold. I assume it's so fast that this is fine? And I assume that's when you removed the "Calculating Prior Variances" tqdm feedback?

hyanwong:
Also, for very large inferences, if e.g. N=1e6 and the user has set ` approx_priors=False`, is it worth outputting some sort of progress bar? Perhaps we only display a progress bar for this stage when progress=True and approx_priors=False?

nspope:
Hmm, I removed the progress bar because the variance calculation is done recursively in numba -- so, no way to output progress without slowing things down dramatically.

tests/test_functions.py
nspope: I lean towards keeping them where they are, because they're so simple (and at some point, there's diminishing returns from writing microscopic unit tests). But if you'd rather move them, that's fine too!

hyanwong:
> Hmm, I removed the progress bar because the variance calculation is done recursively in numba -- so, no way to output progress without slowing things down dramatically.

Oh, well, if there's no obvious trick there then just drop it. 

We could output a logging warning that this might take a while if `approx_priors=False` and N is so large that it will take of the order of an hour or more?

nspope:
Yes, that's a good idea

tests/test_functions.py
nspope: sure, i'll move them back in

tests/test_functions.py
hyanwong: Sure, we could simply dump them in a class at the end of that file, right?

nspope:
Ok I think this is done-- except I'm not sure about the default for using the approximate prior. I suppose that if there's missing data, the calculation has to repeated for every realized # of tips across the tree sequence? So, I set the default to 10,000 which should take on the order of a second according to benchmark above. Seem OK?

hyanwong:
A second sounds OK to me, as doe a default of 10,000, so I'll just merge this. Thanks a lot Nate!

PR #309, opened 2023-08-17
Fix excess mutation count at last edge (closes #308)

### LINKED ISSUE #308: Example of incorrect mutation count in `Likelihoods`
Benchmarking on simulated+inferred tree sequences (example: [inferred.zip](https://github.com/tskit-dev/tsdate/files/12373854/inferred.zip)), there are occasionally very large outliers:

Looking at the EP updates for this node, one of its incoming edges has been allocated a very large number (218) of mutations. For example,
```
import tskit
import tsdate
import numpy as np
ts = tskit.load("inferred.trees")
mut_edges = tsdate.core.VariationalLikelihoods(ts).mut_edges
# array([  1,   1,   0, ...,   0,   1, 218])
```

But this appears to be incorrect, because in the tree sequence only a single mutation is allocated to this edge:
```
outlier_edge = np.where(mut_edges == 218)[0][0] # edge 106980
print([m for m in ts.mutations() if m.edge == outlier_edge])
# [Mutation(id=181693, site=181469, node=22917, derived_state='T', parent=-1, metadata=b'', time=nan, edge=106980)]
```

It seems fishy that the problematic edge here is the last in the array. `VariationalLikelihoods` is inheriting from `Likelihoods` so my guess is the problem lies somewhere in there.

nspope:
Fix minor bug where unmapped mutations are assigned to the last edge by `Likelihoods`, causing a single large outlier when dating

### DIFF
--- a/tsdate/core.py
+++ b/tsdate/core.py
@@ -130,7 +130,8 @@ class Likelihoods:
         """
         mut_edges = np.zeros(ts.num_edges, dtype=np.int64)
         for m in ts.mutations():
-            mut_edges[m.edge] += 1
+            if m.edge != tskit.NULL:
+                mut_edges[m.edge] += 1
         return mut_edges
 
     @staticmethod


hyanwong:
LGTM. We can add a test later.

Issue #301, opened 2023-07-22
Comparing dates between simulated and inferred tree sequences


nspope:
For tree sequences inferred from simulations, we can't directly compare inferred dates to true dates because node sets differ between the inference and the simulation. 

Instead, we've been looking at summary statistics such as the distribution of pairwise coalescence times, relative to what is expected under the coalescent with recombination. Unfortunately, time-indexed statistics are likely to be biased if there are polytomies in the inferred tree sequence. It's not yet clear to me how to systematically characterize/correct for these biases, so as to assess whether or not the inferred tree sequence has reasonable estimates for node dates. For example, polytomies may contribute to the distortions seen in https://github.com/tskit-dev/tsdate/issues/198#issuecomment-1631224197. This seems especially problematic for comparing tsdate to other inference tools like Relate that only output binary trees.

@petrelharp and @hfr1tz3 pointed me towards a tree discrepancy metric they used in https://github.com/petrelharp/num_edges, that gives a meaningful way to compare ages across different tree sequences by finding a mapping between nodes. For a given node in the inferred tree sequence, the idea is to find the node in the true tree sequence that shares the same set of descendants over the longest span. This is something like finding the "true" ancestral segment that is most similar to a given inferred ancestral segment.

This seems to work quite well on a small (1Mb, 200 samples) example, plotting the inferred ages against the ages of the "most similar" nodes from the simulation:

<img src="https://github.com/tskit-dev/tsdate/assets/5767783/1710f112-b9cc-4c0c-a77b-6f0a83aef222" width="50%">

and suggests that tsdate is doing better than we'd judge from the global statistics in https://github.com/tskit-dev/tsdate/issues/198#issuecomment-1631224197. It'd be great to make this fast enough to work with larger simulations. Code:

```python
import msprime
import tsinfer
import tsdate
import tskit
import numpy as np
import matplotlib.pyplot as plt

def match_node(node, inferred_ts, true_ts):
    """
    For `node` in `inferred_ts`, return the age of the node in `true_ts` 
    that has the longest span with the same set of descendant samples
    as `node`. From:

    https://github.com/petrelharp/num_edges/blob/main/Discrepancy%20Function.ipynb
    """
    shared_span = np.zeros(true_ts.num_nodes)
    node_span = 0.0
    for interval, t1, t2 in inferred_ts.coiterate(true_ts):
        desc_set = set(t1.samples(node))
        if len(desc_set) > 0:
            tree_span = interval.span
[...]

PR #310, opened 2023-08-18
Method for matching node sets across tree sequences

nspope:
#301 describes a method for matching nodes between two tree sequences based on the span where both nodes subtend the same sets of samples, based off an implementation by @petrelharp and @hfr1tz3 for evaluating extend_edges. This only uses graph structure, not mutations, so should be useful for comparing tsinfer to other inference pipelines.

Here I've implemented an efficient algorithm to do this:
- If we have a dict mapping from unique sample sets to nodes for each overlapping tree in the query and target tree sequences, it's easy to check if two nodes subtend exactly the same set of samples
- The `evaluation.CladeMap` class is a `tskit.Tree`-like iterator that maintains/updates this dict via edge diffs
- `evaluation.CladeMap.next` returns changes (in terms of unique sample sets) from one tree to the next
- So, when coiterating over the two tree sequences, we only need to update the shared spans for nodes with an altered sample set
This is reasonably quick: for example it takes ~10 seconds to calculate shared spans between a 27000-tree inferred TS (~30000 nodes) and a 100000-tree simulated TS (~80000 nodes), each with 500 samples.

I put this into the new `tsdate.evaluation` module, later I'll add some other utilities to compare ages across tree sequences here at some point (like @petrelharp's mutation mapping strategy). I think we'll also want to add some tools for downsampling polytomies to get unbiased marginal statistics.


nspope:
Here's an example showing node ages in an inferred+dated tree sequence vs their best-matching counterparts in the true ARG. This is 500 samples across 50Mb in stdpopsim's BosTau model. Using the variational dating algorithm,



and the same example using the old inside-outside algorithm,

tests/test_evaluation.py
```diff
@@ -0,0 +1,141 @@
+# MIT License
+#
+# Copyright (c) 2021-23 Tskit Developers
+# Copyright (c) 2020-21 University of Oxford
```
petrelharp, line 4:
```suggestion
```
I think since this file didn't exist previously we don't need this line?

tests/test_evaluation.py
```diff
@@ -0,0 +1,141 @@
...
+
+# --- simulate test case ---
+true_ts = msprime.sim_ancestry(
+    samples=10,
+    sequence_length=1e5,
+    population_size=1e4,
+    recombination_rate=1e-8,
+    random_seed=1024,
+)
+true_ts = msprime.sim_mutations(true_ts, rate=2e-8, random_seed=1024)
```
petrelharp, line 45: perhaps assert that the example has >1 tree? (I see that it ought to, but to make sure...)

tests/test_evaluation.py
```diff
@@ -0,0 +1,141 @@
...
+        clade_to_node = defaultdict(set)
+        for node in tree.nodes():
+            clade = frozenset(tree.samples(node))
+            clade_to_node[clade].add(node)
+        return clade_to_node
+
+    assert ts.sequence_length == other.sequence_length
+    assert ts.num_samples == other.num_samples
+    query_tree = ts.first()
+    target_tree = other.first()
```
petrelharp, line 66: you could do
```
for (interval, query_tree, target_tree) in ts.coiterate(other):
    left, right = interval
```
and cut out a bunch of this.

tsdate/evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        nodes = set()
+        node = edge.parent
+        clade = self._clades[edge.child]
+        while node != tskit.NULL:
+            last = self._clades[node]
+            self._clades[node] = last - clade if downdate else last | clade
+            if len(last):
+                self._nodes[last].remove(node)
+                if not len(self._nodes[last]):
```
petrelharp, line 71:
```suggestion
                if len(self._nodes[last]) > 0:
```

tests/test_evaluation.py
```diff
@@ -0,0 +1,141 @@
...
+                prev = frozenset(tree_1.samples(n))
+                curr = frozenset(tree_2.samples(n))
+                if prev != curr:
+                    diff_test[n] = (prev, curr)
+            for node in diff_test.keys() | diff.keys():
+                assert diff_test[node][0] == diff[node][0]
+                assert diff_test[node][1] == diff[node][1]
+            if tree_2.index == ts.num_trees - 1:
+                break
+            tree_1.next()
```
petrelharp, line 123: It would be good to have a test example where more than one node has the same set of subtended samples? (I think an msprime.CensusEvent would produce this?)

tsdate/evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+    nodes (one from `ts`, one from `other`) subtend the same set of samples.
+
+    Returns two vectors: the age of the best matching node in `other` (e.g.
+    with the longest shared span); and the proportion of the node span in `ts`
+    that is covered by the best match.
+    """
+
+    shared_spans = shared_node_spans(ts, other)
+    matched_span = shared_spans.max(axis=1).todense().A1
+    best_match = shared_spans.argmax(axis=1).A1
```
petrelharp, line 244: note that this will have spurious values of `0` in it that ought to be replaced by `tskit.NULL` (or something?) for other uses - maybe do that here, too, as the code will be copied?

tests/test_evaluation.py
petrelharp: A nice way of testing things would be to run the simulation with a bunch of CensusEvents, to insert unary nodes, then check that when you compare a simplified version of this tree sequence to the original one, that everything matches exactly (eg times). And that the right thing happens when comparing the other way.

petrelharp (approved):
Gee, this is excellent. A few comments.

tsdate/evaluation.py
nspope: these are set to np.nan a couple lines below -- I think that's fine?

nspope:
Random thought -- I've been using RMSE of log10 age to measure estimation error, but this really isn't ideal, as it's most sensitive to the youngest nodes. A better way to calculate goodness-of-fit might be take the mean and variance of the variational distribution, call these $\tilde{\mu_i}$ and $\tilde{\sigma_i}^2$ for the $i$th node, and transform to z-scores $(t_i - \tilde{\mu}_i) \tilde{sigma}_i^{-1}$ before averaging. This would weight the "residual" for each node by how accurately we think we've estimated its age.

tsdate/evaluation.py
petrelharp: the code is fine - I was just thinking ahead to wanting to return the best_match itself

nspope:
Darn-- the shared span calculation works just fine for unary nodes, but there can be multiple "best matches" for a given node (e.g. that all have the same shared span). In particular this can happen with census events. For now, I'll just document what happens in this case (the node with the lowest id is returned). Everything works as expected with simplified tree sequences, and dating with unary nodes is a work-in-progress anyway.

nspope:
OK, I think this is good to go as a first pass.

petrelharp:
>  there can be multiple "best matches" for a given node 

Right - I think that's unavoidable - but won't happen in the limit of large mutation rate? (Or is there something else going on I'm not thinking of?)

hyanwong:
This all seems good to me, thanks for finding an efficient method @nspope. I wonder, however, if it is something that is more widely useful than just for tsdate? There is a `tscompare` repo (https://github.com/tskit-dev/tscompare) with nothing in it, and I wonder if we put this in there, and make it a dependency of tsdate?

Re logging, is there an argument that we should square root the times? This would also allow zeros without the +1 hack, of course. But there's not specific theoretical justification for this (although the expected variance is the square of the mean tMRCA, right?)

hyanwong:
> > I've been using RMSE of log10 age
> 
> Using the posterior itself seems iffy, as you say. What if we just try to ad-hoc visually fit some functional form to the plot of mean error versus age? (and then maybe post hoc figure out a theoretical reason?)

Since theory predicts that the variance in tMRCA goes as `mean^2`, could we simply weight the RMSE by `1/(mean^2 + const)`, where `const` accounts for the fact that even at time zero we have some error in node time estimation? I don't know what the constant should be, however.

nspope:
I'm all for moving this into tscompare (or anywhere else it might be useful), once we've played around with it a bit.

PR #314, opened 2023-09-07
Update packaging

benjeffery:
Bring packaging in line with other tskit repos - I've also made the changes for PyPI uploading but this won't work until I get permissions at PyPI.


.circleci/config.yml
```diff
@@ -44,6 +44,29 @@ jobs:
...
+            rm dist/*
+            python -m build
+
+      - run:
+          name: Install from the distribution tarball
+          command: |
+            python -m venv venv
+            source venv/bin/activate
+            pip install dist/*.tar.gz
+            python -c 'import tsinfer; print(tsdate.__version__)'
```
jeromekelleher, line 64: tsinfer, here and below

.github/workflows/docker/buildwheel.sh
```diff
@@ -13,9 +13,12 @@ git config --global --add safe.directory /project
 # Fetch the full history as we'll be missing tags otherwise.
 git fetch --unshallow
```
jeromekelleher, line 14: tsdate is Python only right, so can we ditch the docker stuff entirely?

jeromekelleher (commented):
LGTM - let's ditch the docker manylinux build if we can though.

.circleci/config.yml
benjeffery: Fixed

.github/workflows/docker/buildwheel.sh
benjeffery: Yep, removed.

benjeffery:
Think this is good to go!

benjeffery:
@hyanwong Can I get a review here? Would like to merge if poss.

hyanwong (approved):
LGTM. Not sure where the `author_email` and `description` etc fields go now, but assume it's all fixed/

benjeffery:
Yikes, thanks for that, I hadn't added `setup.cfg` to the PR! Should be there now.

PR #317, opened 2023-09-22
Discrepancy Function 

hfr1tz3:
This discrepancy functions gives us a metric for comparing two tree sequences. 
The `tree_discrepancy` method returns for the discrepancy between tsa and tsb, the tuple of
a. the total shared span divided by total node span in tsa (which we need to compute); so this is proprotion of the span in tsa that is accurately represented in tsb; and
b. the root-mean-squared discrepancy in time, with the average weighted by span in tsa.


nspope:
Awesome, let me know when you want feedback.

hfr1tz3:
So we need to pass over edges and pass over roots. To do so we should write a separate function. 
Proposal: Need to get per node total span in ts ( ie. ) w_i  in sum (t_x-t_y)**2 w_i/ sum (w_i)
count up in each node the total span from where it is a child which we can get from the edge table np.bincount...
doesn't account for nodes which are roots so we need to include contribution of nodes who are roots.

petrelharp:
```

def node_spans(ts):
    """
    Returns the array of "node spans", i.e., the `j`th entry gives
    the total span over which node `j` is in the tree (i.e., does
    not have 'missing data' there).
    """
    child_spans = np.bincount(
            ts.edges_child,
            weights=ts.edges_right - ts.edges_left,
            minlength=ts.num_nodes,
    )

    for t in ts.trees():
        span = t.interval[1] - t.interval[0]
        for r in t.roots:
            child_spans[r] += span
```

hfr1tz3:
I think that is everything we need for `tree_discrepancy`. I think I will be ready for feedback now @nspope .

nspope:
Oh sorry I just saw this! Will look ASAP.

hfr1tz3:
I still haven't checked if the tests from `test_evaluation.py` have passed yet, but I am working on trying to do that.

petrelharp:
Hold off @nspope - we looked at this today and @hfr1tz3 has homework yet.

hfr1tz3:
All tests in `test_evaluation.py` for `tree_discrepancy` now pass!
@petrelharp does this mean we are ready for review?

tests/test_evaluation.py
```diff
@@ -82,6 +85,55 @@ def _clade_dict(tree):
...
+        time_discrepancies[i] = time_array[i, j]
+    node_span = naive_node_span(ts)
+    total_node_spans = np.sum(node_span)
+    discrepancy = 1 - np.sum(best_match_spans) / total_node_spans
+    rmse = np.sqrt(np.sum(node_span * time_discrepancies**2) / total_node_spans)
+    print("best_match", best_match_spans)
+    print("node_span", node_span)
+    print("shared_spans", shared_spans)
+    print("match", match)
+    print("time_array", time_array)
```
petrelharp, line 133:
```suggestion
```
stray print statements

tests/test_evaluation.py
```diff
@@ -145,3 +206,138 @@ def test_match_self(self, ts):
...
+        check_dis, check_err = naive_discrepancy(pair[0], pair[1])
+        test_dis, test_err = evaluation.tree_discrepancy(pair[0], pair[1])
+        assert np.isclose(check_dis, test_dis)
+        assert np.isclose(check_err, test_err)
+
+    @pytest.mark.parametrize(
+        "pair",
+        [(true_ext, true_ext), (true_simpl, true_ext), (true_simpl, true_unary)],
+    )
+    def test_tree_discrepancy(self, pair):
```
petrelharp, line 228:
```suggestion
    def test_zero_discrepancy(self, pair):
```
This name is more descriptive?

tests/test_evaluation.py
```diff
@@ -145,3 +206,138 @@ def test_match_self(self, ts):
...
+        assert np.isclose(dis, 4 / 46)
+        assert err == 0
+
+    def test_discrepancy_error(self):
+        ts = self.get_simple_ts()
+        other = self.get_simple_ts(time=True)
+        dis, err = evaluation.tree_discrepancy(ts, other)
+        true_error = np.sqrt((2 * 6 * 300**2 + 2 * 2 * 150**2) / 46)
+        assert dis == 0.0
+        assert np.isclose(err, true_error)
```
petrelharp, line 343: Why not add another test that tests with both time and span True?

tsdate/evaluation.py
```diff
@@ -252,3 +252,107 @@ def match_node_ages(ts, other):
...
+            # do this check to exempt 'missing data'
+            if t.num_children(r) > 0:
+                child_spans[r] += span
+    return child_spans
+
+
+def tree_discrepancy(ts, other):
+    """
+    For two tree sequences `ts` and `other`,
+    this method `tree_discrepancy` returns two values:
```
petrelharp, line 280:
```suggestion
    this method returns two values:
```

tsdate/evaluation.py
```diff
@@ -252,3 +252,107 @@ def match_node_ages(ts, other):
...
+    1. the sum across the nodes of `ts`
+    of the length of the discrepancy in span between the node
+    and its best match node in `other`
+    weighted by difference in time.
```
petrelharp, line 284:
```suggestion
    1. the fraction of the total span of `ts` over which each nodes' descendant
    sample set does not match its' best match's descendant sample set.
```
and perhaps say later on
```
    The first value is computed as `A/B`, where `B` is the sum of the spans
    of all nodes in `ts`, and `A` is the
    sum across the nodes of `ts` of the length of the span
    over which the node and its best match node in `other` do not have the
    same set of descendant samples
    weighted by difference in time.
```

tsdate/evaluation.py
```diff
@@ -252,3 +252,107 @@ def match_node_ages(ts, other):
...
+    2. The root mean squared difference
+    between the times in the nodes in `ts`
+    and times of their best matching nodes in `other`
+    with the average weighted by the span in `ts`.
```
petrelharp, line 288:
```suggestion
    2. The root mean squared difference
    between the times of the nodes in `ts`
    and times of their best matching nodes in `other`,
    with the average weighted by the nodes' spans in `ts`.
```

tsdate/evaluation.py
```diff
@@ -252,3 +252,107 @@ def match_node_ages(ts, other):
...
+
+    This is done as follows:
+
+    For each node in `ts` the best matching node(s) from `other`
+    has the longest matching span using `shared_node_spans`.
+    If either tree sequence contains unary nodes there may be
+    multiple matches with the same longest shared span
+    for a single node.
+    In this case, the best match is the node closest in time.
```
petrelharp, line 297:
```suggestion
    If there are multiple matches with the same longest shared span
    for a single node, the best match is the match that is closest in time.
```
There could be multiple matches without unary nodes if there are non-sample leaves.

petrelharp:
I have some minor suggestions, then yes - ready for review!

tests/test_evaluation.py
hfr1tz3: Now that you mention it, that would make sense to do.

petrelharp:
Have a look, @nspope ?

tests/test_evaluation.py
```diff
@@ -1,147 +1,344 @@
...
+    assert len(max_span) == ts.num_nodes
+    match = np.zeros((ts.num_nodes, other.num_nodes))
+    time_array = np.zeros((ts.num_nodes, other.num_nodes))
+    discrepancy = np.zeros((ts.num_nodes, other.num_nodes))
+    for i in range(ts.num_nodes):
+        for j in range(other.num_nodes):
+            if shared_spans[i, j] == max_span[i]:
+                match[i, j] = shared_spans[i, j]
+                time_array[i, j] = np.abs(ts.nodes_time[i] - other.nodes_time[j])
+                discrepancy[i, j] = 1 / (1 + match[i, j] * time_array[i, j])
```
nspope, line 118: Can you remind me of the rationale here? 

The best match should be the node that has the greatest shared span, if this is unambiguous. If it *is* ambiguous (nodes are tied in max shared span), then we could choose the best match to be the tied node with with the closest age to the focal node.

But that two-step procedure isn't happening here, right? So we could conceivably get a "best-matching" node with a relatively small shared span but a very similar age?

tsdate/evaluation.py
```diff
@@ -1,254 +1,354 @@
...
+        shape=(ts.num_nodes, other.num_nodes),
+    )
+    # scale with difference in node times
+    # determine best matches with the best_match_matrix
+    ts_times = ts.nodes_time[row_ind[match]]
+    other_times = other.nodes_time[col_ind[match]]
+    time_difference = np.absolute(np.asarray(ts_times - other_times))
+    best_match_matrix = scipy.sparse.coo_matrix(
+        (
+            1 / (1 + (match_matrix.data * time_difference)),
```
nspope, line 331: Same as earlier comment -- couldn't we end up with best matches that don't satisfy the "max shared span" criterion, but are very similar in terms of age?

And, the choice of time scaling would impact the chance of this happening, right? In that, if the time scaling is such that the absolute difference between times is large relative to typical node spans, then this one-step criterion would be dominated by the difference in ages?

tsdate/evaluation.py
```diff
@@ -1,254 +1,354 @@
...
+    between the times of the nodes in `ts`
+    and times of their best matching nodes in `other`,
+    with the average weighted by the nodes' spans in `ts`.
+
+    This is done as follows:
+
+    For each node in `ts` the best matching node(s) from `other`
+    has the longest matching span using `shared_node_spans`.
+    If there are multiple matches with the same longest shared span
+    for a single node, the best match is the match that is closest in time.
```
nspope, line 293: This agrees with the two-step procedure I wrote in earlier comment, but I don't think that's exactly what's happening below?

nspope (commented):
Looks good, thanks! Just have a couple questions about the "one step" procedure to find the best matching node.

tests/test_evaluation.py
nspope: Oh, I see-- you are doing the two step procedure, sorry! What threw me off was `discrepancy[i, j] = 1 / (1 + match[i, j] * time_array[i, j])` ... does this need to include `match`?

tsdate/evaluation.py
```diff
@@ -1,254 +1,354 @@
...
+    max_span = shared_spans.max(axis=1).toarray().flatten()
+    col_ind = shared_spans.indices
+    row_ind = np.repeat(
+        np.arange(shared_spans.shape[0]), repeats=np.diff(shared_spans.indptr)
+    )
+
+    match = shared_spans.data == max_span[row_ind]
+    # Construct a matrix of potiential matches and
+    match_matrix = scipy.sparse.coo_matrix(
+        (shared_spans.data[match], (row_ind[match], col_ind[match])),
```
nspope, line 321: This would work as intended if `shared_spans.data[match]` was `np.ones(len(match))`, correct?

tsdate/evaluation.py
nspope: Nevermind, I see how this works-- very nice.

tests/test_evaluation.py
```diff
@@ -1,147 +1,344 @@
...
+
+def naive_discrepancy(ts, other):
+    """
+    Ineffiecient but transparent function to compute discrepancy
+    and root-mean-square-error between two tree sequences.
+    """
+    shared_spans = naive_shared_node_spans(ts, other).toarray()
+    max_span = np.max(shared_spans, axis=1)
+    assert len(max_span) == ts.num_nodes
+    match = np.zeros((ts.num_nodes, other.num_nodes))
```
nspope, line 110: is `match` needed? Seems like this would work fine with
```
if shared_spans[i, j] == max_span[i]:
  time_array[i, j] = np.abs(...)
  discrepancy[i, j] = 1 / (1 + time_array[i, j])
```
(also, maybe add a comment about the necessity of 1 / (1 + x) for the sparse matrix format)

tests/test_evaluation.py
```diff
@@ -1,147 +1,344 @@
...
+    time_array = np.zeros((ts.num_nodes, other.num_nodes))
+    discrepancy = np.zeros((ts.num_nodes, other.num_nodes))
+    for i in range(ts.num_nodes):
+        for j in range(other.num_nodes):
+            if shared_spans[i, j] == max_span[i]:
+                match[i, j] = shared_spans[i, j]
+                time_array[i, j] = np.abs(ts.nodes_time[i] - other.nodes_time[j])
+                discrepancy[i, j] = 1 / (1 + match[i, j] * time_array[i, j])
+    best_match = np.argmax(discrepancy, axis=1)
```
nspope, line 119: I'm nitpicking here, but I think a more naive (e.g. clearer to read) implementation would not use sparse matrix operations -- instead, go over `shared_spans` row by row, find the max, find ties, calculate time differences for those ties, and append the best match to to a list.

nspope:
Nevermind, I misunderstood on my initial read-through. Looks great! 

The only real suggestion I have is to rewrite the naive implementation to be more naive (e.g. clearer on a quick glance). That is, don't use sparse matrix operations, but instead loop over rows of `shared_spans`, find the set of nodes matching the row max with `np.where`, get the age difference for these, and append the node with the min age difference to a list.

petrelharp:
> Nevermind, I misunderstood on my initial read-through. Looks great!

Perhaps a short explanatory comment could clarify how it works - I thought it was wrong last time I looked at it also. =)

tests/test_evaluation.py
hfr1tz3: I suppose not, I could just clean it up and use `shared_spans[i,j]`.

petrelharp:
We've got line ending problems; for the record here's what I did to fix them:
```
sed -i 's/^M$//' tsdate/evaluation.py
sed -i 's/^M$//' tests/test_evaluation.py 
```

hyanwong:
Is this ready for merging? I'm happy to do so if @petrelharp and/or @nspope give the OK (assume it's fine for the moment in the `tsdate` repo).

nspope:
I think this PR is waiting on `extend_edges` for the tests to pass? Which might now be merged into the github head for tskit. But maybe we should wait until this is in a release -- I don't think there's an immediate rush.

petrelharp:
It is merged to tskit, but we need a release.

nspope:
Note: we need to ensure that the discrepancy function and time RMSE are calculated correctly when nodes have no matches. Currently, the tie would be broken by comparing to all nodes in the tree sequence (we think?). @hfr1tz3 will fix

petrelharp:
Note that when we add stuff to a tree sequence we add both right and wrong stuff; if the  proprotion of wrong stuff we add is greater than the proportion of wrong stuff already there, we increase discrepancy, even if this proportion is very low. (Example: extending a true but simplified tree sequence.)

So, proposal is that `tree_discrepancy(a, b)` also returns the proportion of the total span of `b` that is matched in `a`; this would be
```
(1 - discrepancy) * total_span(a) / total_span(b)
```

petrelharp:
Also, a `evaluation.total_span(ts)` function that just does `np.sum(evaluation.node_spans(ts))`.

tsdate/evaluation.py
```diff
@@ -274,16 +274,28 @@ def node_spans(ts):
...
-    1. the fraction of the total span of `ts` over which each nodes' descendant
+    this method returns three values, as a tuple:
+    1. The fraction of the total span of `ts` over which each nodes' descendant
     sample set does not match its' best match's descendant sample set.
     2. The root mean squared difference
     between the times of the nodes in `ts`
     and times of their best matching nodes in `other`,
     with the average weighted by the nodes' spans in `ts`.
+    3. The proportion of the span in `ts` that is correctly
+    represented in `other`.
```
petrelharp, line 298: Hm, I think what we talked about was the other way around: 
```suggestion
    3. The proportion of the span in `other` that is correctly
    represented in `ts`.

```

petrelharp:
I think the code does what we want but the docs had it the other way around? I could be mixed up, though - please check?

tsdate/evaluation.py
```diff
@@ -302,9 +314,10 @@ def tree_discrepancy(ts, other):
 
     where :math: `T` is the sum of spans of all nodes in `ts`.
 
-    Returns two values:
-    `discrepancy` (float) the value computed above.
+    Returns three values:
+    `discrepancy` (float) the value computed above
     `root-mean-squared discrepancy` (float)
+    `proportion of correct span in `ts` matching in `other` (float)
```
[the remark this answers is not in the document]
hfr1tz3: Yes, This should now be correct.

petrelharp:
Okay - if you agree with my suggested changes, then please hit the "commit change" button on them.

petrelharp:
TODO:
- squash commits
- rebase to `main`
- double check things are ready to merge

However, this currently uses `extend_edges`, which we are renaming `extend_haplotypes`; so let's (a) make the change to `extend_haplotypes`; (b) check the tests run locally; (c) mark skip them so this passes; (d) make an issue to un-skip these when a new tskit with `extend_haplotypes` is released.

hyanwong:
Moved to #428

Issue #320, opened 2023-10-05
Bringing checks to the front of execution.


barneyhill:
I've encountered two issues that have caused execution to be terminated many hours in. These checks should be done as soon as possible - I'll move them to the beginning of [tsdate.date](https://github.com/tskit-dev/tsdate/blob/main/tsdate/core.py#L1197).

Errors I encountered:
```

I was a bit forgetful and forgot to simplify before running... Might be worth adding in the unary node check at the beginning prior to finding the node spans.

(tsinfer) 2 dnanexus@job-GZKXQQjJg8JzQ8fgYXPGgfb2:~/ukb/workflow-dask$ tsdate date UKB_EUR-chr21-filter_minimal-truncate-0.4-0.6-1-mm0-processed.trees UKB_EUR-chr21-filter_minimal-truncate-0.4-0.6-1-mm0-processed-dated.trees 10000 -m 1e-8 --method variational_gamma --progress -t 192


Find Node Spans: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 11959/11959 [2:28:52<00:00,  1.34it/s]
TipCount:   0%|                                                                                                                                           | 0/3 [2:28:52<?, ?it/s]
Traceback (most recent call last):
  File "/home/dnanexus/ukb/tsinfer/bin/tsdate", line 8, in <module>
    sys.exit(main())
  File "/home/dnanexus/ukb/tsinfer/lib/python3.9/site-packages/tsdate/__main__.py", line 5, in main
    cli.tsdate_main()
  File "/home/dnanexus/ukb/tsinfer/lib/python3.9/site-packages/tsdate/cli.py", line 220, in tsdate_main
    args.runner(args)
  File "/home/dnanexus/ukb/tsinfer/lib/python3.9/site-packages/tsdate/cli.py", line 190, in run_date
    dated_ts = tsdate.date(
  File "/home/dnanexus/ukb/tsinfer/lib/python3.9/site-packages/tsdate/core.py", line 1010, in date
    tree_sequence, dates, posteriors, timepoints, eps, nds = get_dates(
  File "/home/dnanexus/ukb/tsinfer/lib/python3.9/site-packages/tsdate/core.py", line 1085, in get_dates
    priors = prior.build_grid(
  File "/home/dnanexus/ukb/tsinfer/lib/python3.9/site-packages/tsdate/prior.py", line 1051, in build_grid
    span_data = SpansBySamples(contmpr_ts, progress=progress)
[...]

PR #321, opened 2023-10-05
Add initial unary node check + valid method check

barneyhill:
Changes discussed in https://github.com/tskit-dev/tsdate/issues/320


hyanwong:
Thanks @barneyhill - could you add a quick test or two to get these covered?

hyanwong:
And currently the existing tests are failing because of the change in behaviour, so I guess this need fixing?

barneyhill:
I'm now following Apologies https://tskit.dev/tskit/docs/stable/development.html - thanks for the patience on this pull request I'm fairly new to unit testing, formatting etc. A couple of Q's:

1. Any ideas why this unary check could be causing a large number of tests to fail? It's not clear to me why this is happening
2. For the tests I want to set allow_unary=False and then supply a ts with a unary node from tests/utility_functions.py + test passing an invalid method to date(). Do you know the best location for these two tests?

hyanwong:
> I'm now following Apologies https://tskit.dev/tskit/docs/stable/development.html - thanks for the patience on this pull request 

No problem. A good time to learn!
 
> 1. Any ideas why this unary check could be causing a large number of tests to fail? It's not clear to me why this is happening

Presumably because (a) in some cases you do want to allow unary nodes, right? It's only when you try making a prior with the default settings that it should fail. (b) it is looking for a specific error message about unary nodes, (using a regex) and your new error message doesn't match that.

> 2. For the tests I want to set allow_unary=False and then supply a ts with a unary node from tests/utility_functions.py + test passing an invalid method to date(). Do you know the best location for these two tests?

Maybe somewhere in test_inference.py?

hyanwong:
It's usually better to try to match the old behaviour (i.e. don't change the existing tests, where possible)

barneyhill:
Nice so it looks like I needed to check num_children_array[:-1] as the virtual root is included at the last index. Another issue I've encountered here is that the multiroot test is actually unary:

```
>>> for tree in tests.utility_functions.multiroot().trees():
...     print(tree.draw_text())
...
10  11
 ┃   ┃
 4   5
┏┻┓ ┏┻┓
0 1 2 3

 8   9
 ┃   ┃
 4   5
┏┻┓ ┏┻┓
0 1 2 3
```

So when we check for tsdate correctly raising a multiroot error instead a unary error is raised beforehand causing the test to fail. Is it a good idea to go in and change this multiroot test to not include unary nodes?

hyanwong:
> Nice so it looks like I needed to check num_children_array[:-1] as the virtual root is included at the last index.

Ah yes, if you use the arrays then you'll need to remove the virtual root.

> Another issue I've encountered here is that the multiroot test is actually unary:

Oops. That's a bug in the test suite then.

> Is it a good idea to go in and change this multiroot test to not include unary nodes?

Yes please. Maybe make this as a (small) separate PR which I can merge first?

tsdate/prior.py
```diff
@@ -462,6 +462,14 @@ def __init__(self, tree_sequence, *, progress=False, allow_unary=False):
             lambda: defaultdict(lambda: defaultdict(base.FLOAT_DTYPE))
         )
 
+        if not allow_unary:
+            for tree in self.ts.trees():
+                # final element corresponding to the tree’s virtual_root - don't consider.
+                if any(tree.num_children_array[:-1] == 1):
```
hyanwong, line 468: This LGTM: but do you know if it takes up much time for large tree sequences (I wonder if `np.any` is faster?)?

If it's only a tiny fraction of the time of the total (which I suspect), then it's fine.

hyanwong (approved):
Generally LGTM

tsdate/prior.py
hyanwong: I've just tried it on https://zenodo.org/record/5512994/files/hgdp_tgp_sgdp_high_cov_ancients_chr2_q.dated.trees.tsz. It takes 4 seconds to iterate through the trees, 3 mins 40 to check unaryness using `np.any`, and > 1 hour using `any()`. I guess that's because it's looking through *all* the nodes (including those not in the tree) for *every* tree. That could be billions of operations.

hyanwong: I think the following incremental-type approach seems to be the fastest: it only check those nodes whose children change (i.e. appear as a parent in an edge) for each tree. This takes just under one and a half minutes on my laptop for a 138MB tree sequence, which I think is acceptable? Do you want to check it does the right thing, @barneyhill ?

```python
for tree, ediff in zip(self.ts.trees(), self.ts.edge_diffs()):
    changed = {e.parent for edges in (ediff.edges_out, ediff.edges_in) for e in edges}
    if (tree.num_children_array[list(changed)] == 1).any():
        raise ValueError
```

hyanwong: Opened as a discussion in https://github.com/tskit-dev/tskit/discussions/2865 - note that we want to call these "locally unary" nodes.

hyanwong: Hi @barneyhill - do you want to change this to the `has_locally_unary_nodes()` method in https://github.com/tskit-dev/tskit/discussions/2865, then I can merge this? I want to get a tsdate release out next week.

hyanwong:
Yep, this all looks good to me. Thanks @barneyhill !

PR #324, opened 2023-10-09
Explicitly defined multiroot example tree

barneyhill:
Example is now unary and should fix the issue in https://github.com/tskit-dev/tsdate/pull/321


barneyhill:
@hyanwong does this look good to you?

hyanwong:
Sorry for the slow reply. Yes, this LGTM. Merging, and thanks @barneyhill

PR #337, opened 2023-11-16
Add ref to schweiger2023ultra

hyanwong:


nspope (approved):
LGTM. The methods are ultimately very different under the hood but I guess the overall motivation is similar.

PR #339, opened 2023-11-16
Docs proofread

nspope:
Some suggested changes to the docs wrt the variational gamma method.

### DIFF
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -18,7 +18,7 @@ kernelspec:
 
 # Introduction
 
-`Tsdate` {cite}`wohns2022unified` infers dates for nodes in a genetic genealogy,
+`tsdate` {cite}`wohns2022unified` infers dates for nodes in a genetic genealogy,
 sometimes loosely known as an ancestral recombination graph or ARG
 {cite}`wong2023general`. More precisely, it takes a genealogy in 
 [tree sequence](https://tskit.dev/tutorials/what_is.html) format as an input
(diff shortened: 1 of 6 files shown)


nspope:
Only change to code is to import `tsdate.priors.parameter_grid` as `tsdate.build_parameter_grid` to expose an analogue for `tsdate.build_prior_grid`

nspope:
One comment: "loopy belief propagation" as mentioned here

https://github.com/tskit-dev/tsdate/blob/183363af4a7cc2bdab5710c04f79afc3fa59d30b/docs/methods.md?plain=1#L74

doesn't make much sense to me-- I've always seen LBP used to refer to a method (apply belief propagation to graphs with loops) rather than as a "theoretical problem", per se. So my vote is to remove references to LBP from `methods.md` (I don't think they will be meaningful to most readers, anyway).

nspope:
I'm having issues building the docs locally, so haven't yet checked to see if the links all work.

hyanwong:
@mergifyio rebase

hyanwong:
> One comment: the phrase "loopy belief propagation" as used here
> 
> https://github.com/tskit-dev/tsdate/blob/183363af4a7cc2bdab5710c04f79afc3fa59d30b/docs/methods.md?plain=1#L74
> 
> doesn't make much sense to me-- I've always seen LBP used to refer to a method (apply belief propagation repeatedly to graphs with loops) rather than as a "theoretical problem", per se. So my vote is to remove references to LBP from `methods.md` (I don't think they will be meaningful to most readers, anyway).

Ah, good point about the terminology. I will try to reword. I think we should make the point that the previous methods (without iterative convergence) were not exact, however.

hyanwong:
I'll merge this and fix any refs plus address the LBP stuff in a follow-up PR

PR #346, opened 2023-12-04
Minor fixes for central moment matching

nspope:
This adds (or rather documents) an option `min_kl`, which if set to False matches central moments rather than sufficient statistics. This is around 2-3 times faster, but introduces some upward bias in the ages of very young nodes.

### DIFF
--- a/tsdate/core.py
+++ b/tsdate/core.py
@@ -1696,6 +1696,10 @@ def variational_dates(
         disable=not progress,
     ):
         dynamic_prog.iterate(max_shape=max_shape, min_kl=min_kl, debug=debug)
+    # (DEBUG) check how many messages were skipped
+    skipped = np.sum(np.isnan(dynamic_prog.log_partition))
+    print("Skipped", skipped, "messages")
+    # (END DEBUG)
 
     posterior = priors.clone_with_new_data(
         grid_data=dynamic_prog.posterior[priors.nonfixed_nodes, :]
--- a/tsdate/core.py
+++ b/tsdate/core.py
@@ -1011,9 +1011,19 @@ class ExpectationPropagation(InOutAlgorithms):
         return internal, external
 
     @staticmethod
-    @numba.njit("f8(i4[:, :], f8[:, :], f8[:, :], f8[:, :, :], f8[:], f8[:], f8, b1)")
+    @numba.njit(
+        "f8(i4[:, :], f8[:, :], f8[:, :], f8[:, :, :], f8[:], f8[:], f8, b1, b1)"
+    )
     def propagate(
-        edges, likelihoods, posterior, messages, scale, log_partition, max_shape, min_kl
+        edges,
+        likelihoods,
+        posterior,
+        messages,
+        scale,
+        log_partition,
+        max_shape,
+        min_kl,
+        debug,
     ):
         """
         Update approximating factors for each edge, returning average relative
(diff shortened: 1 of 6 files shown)


nspope:
I renamed the outward facing option to `method_of_moments` with default `False`. It'd be good to put this option in the docs somewhere, maybe with an example showing the upward bias.

For example here's log(age) with fast method of moments versus less-fast matching sufficient statistics:



It's not too bad, and speeds things up noticeably, so could be useful for the large data case (@hyanwong maybe run on GEL data, and plot vs the default algorithm, if you have the time?)

hyanwong:
This now runs on all the tree sequences (including those inferred from real data) that we have thrown at it. So it's probably OK to merge?

However, it is failing CI because it outputs "Skipped 0 messages" when no messages are skipped. I guess we should not print a message when this happens?

I wonder if it would be useful to encode some of the diagnostics, like which excess were skipped, in metadata somehow. I don;t think we want to use edge metadata (because there is a general recommendation not to store metadata on edges), but I wonder if it would make sense to store a list of the skipped edges e.g. on the nodes.

nspope:
We could use logging.info for the total number of messages?

What about an option to return marginal likelihood contribution per edge? I was thinking this might be useful for diagnostics more generally. 

Alternatively, we could have an option that returns the model object (`dynamic_prog` I think it's called internally) which would have all this stuff (but, would be more for debugging purposes)

hyanwong:
> What about an option to return marginal likelihood contribution per edge? I was thinking this might be useful for diagnostics more generally.

We could, I suppose, have an option which, when set, dumps this in the edge metadata? People could delete it from there if it got in the way of further analysis.

nspope:
yes that sounds good. Maybe we hold off for now and do that in a followup, post-release? As I already started working on the correct marginal likelihood calculation in another PR

hyanwong:
A general option `dump_diagnostic_metadata` which dumps diagnostics in the appropriate metadata might be useful, and not just for edges.

The only down side with this is having to deal with what to do if metadata already exists for a table into which additional stuff is added. But we can just check at the start if that's the case and bomb out if not.

nspope:
then, I think we're good to go here. Do we want to set variational_gamma as the default method for this release, or hold off until there's a preprint?

hyanwong:
Also we are not 100% sure of the API for setting options, right? E.g. we might want to pass some of these in via a config dictionary? As long as we don't document all the options we don't need to maintain them!

tsdate/core.py
```diff
@@ -1274,6 +1281,12 @@ def date(
         from the inside algorithm in addition to the dated tree sequence. If
         ``return_posteriors`` is also ``True``, then the marginal likelihood
         will be the last element of the tuple.
+    :param bool method_of_moments: If ``True`` match central moments in variational gamma
```
hyanwong, line 1284: I'm wondering if we should comment out these parameter descriptions for the moment, as well as making sure the global_prior and max_iterations are undocumented, then we can introduce them once we have decided on the best API? E.g. I wonder if the original idea of a dictionary of control parameters is more flexible? It seems a little odd to have some parameters that don't apply to the `inside-outside` method in there too.

Another possibility would be to define `tsdate.date_vgamma` and `tsdate.date_discrete` (or some such) as the base functions, and simply use `tsdate.date` as a wrapper, which passes `**kwargs` to the base functions (and then we document the specific options within the base functions).

I will ask Ben and Jerome what the best approach is here.

hyanwong (approved):
Looks great. Will merge when we have decided on the API.

PR #350, opened 2023-12-21
Refactor/redocument API

nspope:
- Make `get_dates` and `variational_dates` part of the public API, that will return unconstrained node ages and posteriors. Rename `get_dates` to `discretised_dates`. The main method `date` calls one of these functions and inserts constrained dates into the tree sequence.
- Moves docstrings for method specific arguments into `discretised_dates` and `variational_dates`. Shared arguments are documented in `date`.
- In the docstrings, the posteriors from the inside-outside algo are referred to as being over 'time slices'. This isn't really correct, the likelihood is evaluated at fixed time points rather than integrated over a time interval. So have changed docstrings accordingly. (It's a little confusing because the default prior is calculated over intervals from a CDF, but given that arbitrary priors could be used, this change seems like the right choice.) Accordingly, the returned posteriors now have a single key `time` instead of `start_time` and `end_time`.
- Some miscellaneous refactoring/cleanup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@
   `action="count"`, so `-v` turns verbosity to INFO level,
   whereas `-vv` turns verbosity to DEBUG level.
 
+- The `return_posteriors=True` option with `method="inside_outside"`
+  previously returned a dict that included keys `start_time` and `end_time`,
+  giving the impression that the posterior for node age is discretized over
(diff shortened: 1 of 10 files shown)


nspope:
This is ready for a look whenever you have time @hyanwong.

hyanwong (approved):
This mostly LGTM. I have some suggestions, but I will merge and make suggestions in future PRs

PR #349, opened 2023-12-11
Add mixture-of-gammas global prior, updated by EM

nspope:
1. To use the coalescent prior to initialize a mixture with `K` components, pass `prior_mixture_dim=K`
2. To disable the EM updates of the mixture components, pass `em_iterations=0`
3. The "original" global prior for the variational gamma method can be used via `prior_mixture_dim=1, em_iterations=0`
4. The default is to optimize a single component prior (`prior_mixture_dim=1, em_iterations=10`)

PR #351, opened 2024-01-04
Api extras

hyanwong:
Make `tsdate.date()` a wrapper for `tsdate.inside_outside()`, `tsdate.variational_gamma()`, or `tsdate.maximization()`. This allows easier documentation and reuse, and is a follow-up to #350 

Also ban `tsdate.maximization()` from using the `return_posteriors` parameter.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -25,8 +25,11 @@ This page provides formal documentation for the `tsdate` Python API.
 
 ```{eval-rst}
 .. autofunction:: tsdate.date
-.. autofunction:: tsdate.discretised_dates
-.. autofunction:: tsdate.variational_dates
+.. autodata:: tsdate.core.estimation_methods
+   :no-value:
+.. autofunction:: tsdate.inside_outside
+.. autofunction:: tsdate.maximization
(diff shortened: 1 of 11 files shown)


nspope (approved):
LGTM -- this'll break #349 but it should be easy enough to fix. I worry a little bit about tying the output to a `Results` class, because I think we may want to return different things from different methods ultimately (for example, the variational gamma method gives a per-edge "likelihood" that could be useful, and I'm not sure there's an analogue for inside-outside). But, if it comes to that I suppose we can make method-specific `Results` classes that at a minimum contain the 4 fields in the base case.

nspope:
also what do you think about splitting the contents of `core.py` into different files? It's getting a bit big, what with the API, the algorithms, and the likelihood objects?

We don't have to do that now though.

hyanwong:
Thanks so much for looking at this @nspope. Re the "results" class, this is an internal thing that is never exposed to the user (it's just to make out lives easier), so we can rework it later however we like.

I agree about splitting up `core.py`, which we can do later. It may also be that we can place some of the gamma stuff within the VariationalGammaMethod class?

PR #360, opened 2024-01-10
Rescale messages at end of each iteration

nspope:
Rather than maintain separate scaling factors and unconstrained messages across iterations, absorb the scale into the message at the end of each iteration. The scaling terms are essentially responsible for keeping the posterior shape parameters within a reasonable bound. 

If the scale is never reset, it'll go towards 0 as iterations progress while the (unconstrained) message goes towards infinity. This can underflow and trigger an assert, if the algo is run long enough and a node shows up across a ton of updates (e.g. has an exceptionally long span and/or is involved in large polytomies).

### DIFF
--- a/tsdate/core.py
+++ b/tsdate/core.py
@@ -1001,10 +1001,6 @@ class ExpectationPropagation(InOutAlgorithms):
             self.posterior[p] += self.likelihoods[i]
             # self.log_partition[i] += ... # TODO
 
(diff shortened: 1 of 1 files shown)


nspope:
Small bugfix here @hyanwong

hyanwong:
Nice, thanks for solving this

PR #364, opened 2024-01-17
Non-contemporary samples

nspope:
- Make the `ExpectationPropagation` class a bit more stand-alone
- Add argument `constraints` to the class, that are lower/upper age bounds per node. If these are equal, the node age is fixed to a particular timepoint. If these are (0, inf) the node age is unconstrained. 
- Add machinery for updating edges where one of the nodes is fixed.
- Some numerical fixes
- Utilities for cutting up nodes

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -46,7 +46,6 @@ from tsdate.core import InOutAlgorithms
 from tsdate.core import InsideOutsideMethod
 from tsdate.core import Likelihoods
 from tsdate.core import LogLikelihoods
-from tsdate.core import VariationalLikelihoods
 from tsdate.demography import PopulationSizeHistory
 from tsdate.prior import ConditionalCoalescentTimes
 from tsdate.prior import fill_priors
@@ -55,6 +54,8 @@ from tsdate.prior import MixturePrior
 from tsdate.prior import PriorParams
 from tsdate.prior import SpansBySamples
 from tsdate.util import nodes_time_unconstrained
+from tsdate.util import split_disjoint_nodes
+from tsdate.util import split_root_nodes
 
 
 class TestBasicFunctions:
@@ -788,11 +789,6 @@ class TestLikelihoodClass:
 
 class TestVariational:
     # TODO - needs a few more tests in here
-    def test_variational_prob_space(self):
-        ts = utility_functions.two_tree_mutation_ts()
-        lik = VariationalLikelihoods(ts, mutation_rate=None)
-        assert lik.probability_space == base.GAMMA_PAR
-
     def test_variational_nosize(self):
         ts = utility_functions.two_tree_mutation_ts()
         with pytest.raises(ValueError, match="Must specify population size"):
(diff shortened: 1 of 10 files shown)


hyanwong:
Great to tidy this up, thanks @nspope . At some point we also mooted the idea of tidying up base.py / core.py / etc to sort into a more logical set of files. I assume we still think this is a good idea? You probably have a reasonable grasp as to how the whole codebase fits together, so suggestions for helpful reorganisations are welcome.

nspope:
I was thinking that we'd keep the main API in core.py and move the algorithms/support classes into separate files. I've done that with ExpectationPropagation in this PR, as part of a bigger refactoring.

nspope:
As a litmus test, I simulated an example with 100 samples in each of three groups: (A) "focal" contemporary samples; (B) "extra" contemporary samples; (C) ancient samples between 50 and 5000 generations ago. Then:
1. Simplify to (A) and (B) and date, call this A+B
2. Simplify to (A) and (C) and date, call this A+C
3. Simplify to A and extract the subset of mutations that are segregating in A
4. Get dates for these mutations from A+B and A+C, and compare to the truth

We're looking to see if including an additional 100 ancient samples reduces error relative to including an additional 100 contemporary samples; to make the comparison 1-to-1 we use a shared subset of mutations.

__With the focal group only (e.g. 100 contemporary samples):__



__With the focal + extra (e.g. 200 contemporary):__




__With the focal + ancient (e.g. 100 contemporary + 100 ancient), rug at bottom is age of ancients:__



So seems like this works as expected!

nspope:
@hyanwong this is ready for a look. The docs build is failing for some reason (a cell is timing out), not sure why.

tests/test_functions.py
```diff
@@ -2213,3 +2209,99 @@ def test_bad_arguments(self):
...
+            if len(intr) != 1:
+                return True
+
+        return False
+
+    @staticmethod
+    def childset_changes_with_root(ts):
+        """
+        If root nodes are split whenever their children change, the next root
+        root should have the same child set if it has the same ID
```
hyanwong, line 2256: duplicate "root" in docstring

tsdate/util.py
```diff
@@ -319,3 +326,268 @@ def add_sampledata_times(samples, sites_time):
...
+
+
+def split_disjoint_nodes(ts):
+    """
+    For each non-sample node, split regions separated by gaps into distinct
+    nodes.
+
+    Where there are multiple disconnected regions, the leftmost one is assigned
+    the ID of the original node, and the remainder are assigned new node IDs.
+    Population, flags, individual, time, and metadata are all copied into the
```
hyanwong, line 411: I think we aren't copying metadata any more, right? just assigning the original node ID into the metadata?

tsdate/util.py
```diff
@@ -319,3 +326,268 @@ def add_sampledata_times(samples, sites_time):
...
+    tables = ts.dump_tables()
+    tables.nodes.set_columns(
+        flags=tables.nodes.flags[nodes_order],
+        time=tables.nodes.time[nodes_order],
+        population=tables.nodes.population[nodes_order],
+        individual=tables.nodes.individual[nodes_order],
+    )
+    # TODO: copy existing metadata for original nodes
+    # TODO: add new metadata indicating origin for split nodes
+    # TODO: add flag for split nodes
```
hyanwong, line 442: I think we can add this in this PR easily enough?

tsdate/util.py
```diff
@@ -319,3 +326,268 @@ def add_sampledata_times(samples, sites_time):
...
+    mutations_node[nonsegregating] = ts.mutations_node[nonsegregating]
+
+    tables = ts.dump_tables()
+    tables.nodes.set_columns(
+        flags=tables.nodes.flags[nodes_order],
+        time=tables.nodes.time[nodes_order],
+        population=tables.nodes.population[nodes_order],
+        individual=tables.nodes.individual[nodes_order],
+    )
+    # TODO: copy existing metadata for original nodes
```
hyanwong, line 440: can probably remove this comment?

hyanwong (approved):
This generally LGTM. I presume we should add in some extra tests for historical samples (I can do that). What do we do if we a running the inside-outdise algorithm with historical samples (we can have problems with the prior in this case). I guess we should allow a global prior to be specified in the inside-outside method as well as in the vgamma method?

PR #371, opened 2024-02-17
Piecewise uniform prior for timescale

nspope:
Overhaul of variational dating, in particular re-estimating the timescale after expectation propagation


nspope:
@hyanwong I think we should get this in, and add unit tests for normalisation.py in a followup (once that algorithm stabilizes). Testing is a bit of mess at the moment because of the different algorithms / options / CLI, and needs a cleanup overall.

tests/test_functions.py
```diff
@@ -2288,20 +2323,20 @@ def test_split_disjoint_nodes(self):
...
-        ts = msprime.sim_mutations(ts, rate=1e-8, random_seed=1)
-        sample_data = tsinfer.SampleData.from_tree_sequence(ts)
-        inferred_ts = tsinfer.infer(sample_data).simplify()
-        split_ts = split_root_nodes(inferred_ts)
-        split_root_nodes(ts)
-        assert not self.childset_changes_with_root(inferred_ts)
-        assert self.childset_changes_with_root(split_ts)
-        assert split_ts.num_edges > inferred_ts.num_edges
-        assert split_ts.num_nodes > inferred_ts.num_nodes
+    # def test_split_root_nodes(self):
```
hyanwong, line 2326: Maybe worth commenting:

```
# test here removed because better results are obtained by running `split_disjoint_nodes`
# which splits all nodes of this sort, rather than just the roots. Note that the "ultimate ancestor"
# is split by default in tsinfer anyway.
```

tests/test_inference.py
```diff
@@ -406,84 +408,25 @@ class TestVariational:
...
-            ts,
-            mutation_rate=5,
-            method="variational_gamma",
-            priors=grid,
-            prior_mixture_dim=2,
-        )
-
-    def test_bad_arguments(self):
-        ts = utility_functions.two_tree_mutation_ts()
-        with pytest.raises(ValueError, match="Maximum number of EP iterations"):
```
hyanwong, line 456: I think this test is worth leaving in?

tests/test_inference.py
```diff
@@ -406,84 +408,25 @@ class TestVariational:
...
-        with pytest.raises(ValueError, match="must be a nonnegative integer"):
-            tsdate.date(
-                ts,
-                mutation_rate=5,
-                population_size=1,
-                method="variational_gamma",
-                prior_mixture_dim=0.1,
-            )
-
-    def test_match_central_moments(self):
```
hyanwong, line 473: Should we also leave this in?

tests/test_inference.py
```diff
@@ -406,84 +408,25 @@ class TestVariational:
...
-        )
-        ts1 = tsdate.date(
-            ts,
-            mutation_rate=5,
-            population_size=1,
-            method="variational_gamma",
-            match_central_moments=True,
-        )
-        assert np.any(np.not_equal(ts0.nodes_time, ts1.nodes_time))
+    ts = msprime.sim_ancestry(
```
hyanwong, line 411: I think this should be something like:

```python
class TestVariational:
    @pytest.fixture()
    def ts():
        ts = msprime.sim_ancestry( ...
        return ts

    def test_binary(self, ts):
        ...

```

tsdate/core.py
```diff
@@ -957,47 +939,104 @@ def __init__(
...
+            self.constr_iterations = constr_iterations
+
+        if self.prior_grid_func_name is None:
+            if priors is not None:
+                raise ValueError(f"Priors are not used for method {self.name}")
             if Ne is not None:
-                raise ValueError(
-                    "Cannot specify population size if specifying priors "
-                    f"from tsdate.build_{self.prior_grid_func_name}()"
+                raise ValueError(f"Population size is not used for method {self.name}")
```
hyanwong, line 956: For future ref, see https://github.com/tskit-dev/tsdate/issues/376 which will remove the need to do this.

tsdate/core.py
```diff
@@ -957,47 +939,104 @@ def __init__(
...
+        )
+        mutations.time = util.constrain_mutations(
+            self.ts, nodes.time, self.mutations_edge
+        )
         tables.time_units = self.time_units
-        tables.mutations.time = np.full(self.ts.num_mutations, tskit.UNKNOWN_TIME)
-        # Add posterior mean and variance to node metadata
-        if result.posterior_obj is not None:
+        constr_timing -= time.time()
+        logging.info(f"Constrained node ages in {abs(constr_timing)} seconds")
```
hyanwong, line 1006: Maybe more useful to output an (optional) tqdm bar to give feedback, which presumably would remove the need to report times?

tsdate/variational.py
```diff
@@ -563,32 +595,106 @@ def iterate(
...
                 max_shape=max_shape,
                 min_step=min_step,
                 min_kl=min_kl,
+                regularise=regularise,
+            )
+        nodes_timing -= time.time()
+        skipped_nodes = np.sum(np.isnan(self.log_partition))
+        if skipped_nodes:
+            logging.info(f"Skipped {skipped_nodes} nodes with invalid posteriors")
+        logging.info(f"Calculated node posteriors in {abs(nodes_timing)} seconds")
```
hyanwong, line 674: Similarly here - maybe a tqdm progress bar would be better than reporting timings?

hyanwong (approved):
All LGTM. You can ignore as many comments as you like @nspope, or open an issue and we can address anything later. Let me know if you would like me to merge?

evaluation/evaluate_accuracy.py
```diff
@@ -1,813 +0,0 @@
-import argparse
```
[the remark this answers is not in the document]
nspope: That code doesn't run through, and was causing import issues due to another module named evaluation. Could include accuracy benchmarks in a directory "benchmarks" at some point.

tests/test_cli.py
```diff
@@ -248,7 +248,7 @@ def test_verbosity(self, tmp_path, caplog, flag, log_status):
     )
     def test_no_progress(self, method, tmp_path, capfd):
         input_ts = msprime.simulate(4, random_seed=123)
-        params = f"-m 0.1 --method {method}"
+        params = f"-m 0.1 --method {method} --rescaling-intervals 0"
```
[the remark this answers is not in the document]
nspope: yeah this can be removed at a later day when the "too many intervals" issue is fixed

tsdate/core.py
nspope: The constraint is all done in numba so no tqdm bar. I'd prefer to keep logging things as I do here.

tsdate/variational.py
nspope: I prefer logging, not a fan of progress bars, and they won't work with numba anyway

tests/test_inference.py
nspope: agreed, done

tsdate/core.py
hyanwong: Fair enough.

tests/test_inference.py
nspope: I think it's fine -- I plan on removing this option as it just causes numerical instability with no improvement in accuracy.

evaluation/evaluate_accuracy.py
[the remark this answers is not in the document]
hyanwong: Yep: I think we can simply keep that PR open until we implement something similar under a different name, as you say, something like "accuracy_benchmarks".

tests/test_inference.py
nspope: ... to clarify, we should add tests for various options, but right now i'm swapping them in and out enough that it's a pain. So stuff like this'll get hit in the testing overhaul

tests/test_functions.py
nspope: hm, the splitting code is no longer in tsdate, so I don't think it makes sense to have a comment. I'll just remove the test.

nspope:
Thanks for looking @hyanwong, let's merge this in

hyanwong:
Done!

PR #382, opened 2024-05-30
Save flags and metadata when splitting disjoint nodes

### LINKED ISSUE #373: Add `split_disjoint_nodes` to preprocess routine
It looks like we are defaulting to running @nspope's `utils.split_disjoint_nodes()` before dating, certainly for the variational gamma method. We should add this to `tsdate.preprocess_ts` (performed *after* the simplification) and also make it a command-line option, e.g. here:

https://github.com/tskit-dev/tsdate/blob/9490955a9e62391ccdff0418c3d75a22c3e94205/tsdate/cli.py#L307

I guess we default to doing this, if we want most people to use it?

hyanwong:
And add split_nodes to preprocess_ts. Fixes #373

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -28,31 +28,85 @@ import logging
 import msprime
 import numpy as np
 import pytest
+import tsinfer
 import tskit
 
 import tsdate
 
 
 class TestSplitDisjointNodes:
+    @staticmethod
+    def has_disjoint_nodes(ts):
+        """
+        Brute force check for disjoint nodes, by pulling out edge intervals for
+        each node; taking the union of intervals; checking that a single
+        interval remains
+        """
+
+        def merge_intervals(intervals):
+            intervals = sorted(intervals, key=lambda x: x[0])
+            result = []
+            (start_candidate, stop_candidate) = intervals[0]
+            for start, stop in intervals[1:]:
+                if start <= stop_candidate:
+                    stop_candidate = max(stop, stop_candidate)
+                else:
+                    result.append((start_candidate, stop_candidate))
+                    (start_candidate, stop_candidate) = (start, stop)
+            result.append((start_candidate, stop_candidate))
+            return result
+
+        intervals_by_node = {i: [] for i in range(ts.num_nodes)}
+        for e in ts.edges():
+            intervals_by_node[e.parent].append([e.left, e.right])
+            intervals_by_node[e.child].append([e.left, e.right])
+
+        for n in range(ts.num_nodes):
+            intr = merge_intervals(intervals_by_node[n])
+            if len(intr) != 1:
+                return True
+
+        return False
+
+    @staticmethod
+    def childset_changes_with_root(ts):
+        """
+        If root nodes are split whenever their children change, the next root
+        should have the same child set if it has the same ID
+        """
+        last_childset = frozenset()
+        last_root = tskit.NULL
+        for t in ts.trees():
+            if t.num_edges == 0:
+                last_childset = frozenset()
+                last_root = tskit.NULL
+            else:
+                if t.num_roots > 1:
+                    return False
+                childset = frozenset(list(t.children(t.root)))
+                if t.root == last_root and childset != last_childset:
+                    return False
+                last_childset = childset
+                last_root = t.root
+        return True
+
     def test_nosplit(self):
         ts = tskit.Tree.generate_comb(5).tree_sequence
         split_ts = tsdate.util.split_disjoint_nodes(ts)
-        assert split_ts.num_nodes == ts.num_nodes
-        assert split_ts.num_edges == ts.num_edges
-        assert split_ts.num_trees == ts.num_trees
-        for node in split_ts.nodes():
-            assert node.flags & tsdate.NODE_SPLIT_BY_PREPROCESS == 0
+        assert ts.equals(split_ts, ignore_provenance=True)
         prov = json.loads(split_ts.provenance(-1).record)
         assert prov["software"]["name"] == "tsdate"
         assert prov["parameters"]["command"] == "split_disjoint_nodes"
 
-    def test_simple(self):
+    def test_simple(self, caplog):
         tables = tskit.Tree.generate_comb(5).tree_sequence.dump_tables()
         tables.delete_intervals([[0.2, 0.8]])
         tables.nodes.metadata_schema = tskit.MetadataSchema.permissive_json()
         ts = tables.tree_sequence()
         num_internal_nodes = ts.num_nodes - ts.num_samples
-        split_ts = tsdate.util.split_disjoint_nodes(ts)
+        with caplog.at_level(logging.WARNING):
+            split_ts = tsdate.util.split_disjoint_nodes(ts)
+            assert caplog.text == ""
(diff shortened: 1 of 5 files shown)


hyanwong:
@nspope : I also added the following to the changelog. Do my short summaries read OK?

```
- Variational gamma uses a rescaling approach which helps considerably if e.g.
  population sizes vary over time

- Variational gamma does not use mutational area of branches, but average path
  length, which reduces bias in tree sequences containing polytomies

**Breaking changes**

- Variational gamma uses an improper (flat) prior, and therefore
  no longer needs `population_size` specifying.

```

hyanwong:
Also note the API for adding metadata information on split nodes: `split_disjoint_nodes(ts, metadata_key="original_node_id")`. If metadata_key is of non-zero-length, then original node ID is added to the split nodes using this key. By default the key is `None` treated as `""` - meaning "don't try to add any metadata". This means the routine will work (and not include extra metadata) for any metadata schema.

I'm unsure if this seems a bit hacky? But it seems good to me to always be able to run the routine by default.

nspope:
> do my short summaries read OK?

LGTM! Let me know when you're ready for a review.

tsdate/util.py
```diff
@@ -344,9 +370,48 @@ def mutation_span_array(tree_sequence):
...
 
 
-@numba.njit(_unituple(_i1w, 3)(_i1r, _i1r, _f1r, _f1r, _b1r))
+# Some functions for changing tskit metadata
+# See https://github.com/tskit-dev/tskit/discussions/2954
+def _change_ragged_order(ragged_arr, ragged_offset, new_order):
+    # returns a tuple of a new ragged_col in a new order and a new offset array
+    old_ranges = np.array([ragged_offset[:-1], ragged_offset[1:]]).T
+    ranges = old_ranges[new_order, :]
+    ix = [np.arange(lf, rt, dtype=ragged_offset.dtype) for lf, rt in ranges if lf != rt]
```
nspope, line 379: Because of this comprehension, this function seems like a good candidate for numba-fication. Have you benchmarked this routine on very large treeseqs, versus the tskit.unpack_strings approach?

hyanwong:
Definitely ready for a review, thanks!

Re your comment about `tskit.unpack_strings`, that's a good point, and I haven't tried this, no.

tsdate/util.py
hyanwong: Good point. I guess there are 2 scenarios

1. large ts with no node metadata (easy e.g. to simulate)
2. large ts with metadata (perhaps one of the unified genealogy files, e.g. https://zenodo.org/records/5512994/files/hgdp_tgp_sgdp_high_cov_ancients_chr2_q.dated.trees.tsz)

hyanwong:
```python
from datetime import datetime
import tszip
ts = tszip.decompress("hgdp_tgp_sgdp_high_cov_ancients_chr2_q.dated.trees.tsz")

def _change_ragged_order(ragged_arr, ragged_offset, new_order):
    # returns a tuple of a new ragged_col in a new order and a new offset array
    old_ranges = np.array([ragged_offset[:-1], ragged_offset[1:]]).T
    ranges = old_ranges[new_order, :]
    ix = [np.arange(lf, rt, dtype=ragged_offset.dtype) for lf, rt in ranges if lf != rt]
    select = [] if len(ix) == 0 else np.concatenate(ix)
    return ragged_arr[select], np.insert(np.cumsum(np.diff(ranges, axis=1)), 0, 0)

def _change_ragged_order2(ragged_arr, ragged_offset, new_order):
    # returns a tuple of a new ragged_col in a new order and a new offset array
    arr = tskit.unpack_arrays(ragged_arr, ragged_offset)
    return tskit.pack_arrays([arr[i] for i in new_order])

def _change_ragged_order3(ragged_arr, ragged_offset, new_order):
    # returns a tuple of a new ragged_col in a new order and a new offset array
    arr = tskit.unpack_bytes(ragged_arr, ragged_offset)
    return tskit.pack_bytes([arr[i] for i in new_order])


nodes = ts.tables.nodes
start = datetime.now()
a = _change_ragged_order3(nodes.metadata, nodes.metadata_offset, ts.num_nodes - np.arange(ts.num_nodes) - 1)
print(f"took {datetime.now() - start} secs")
```

hyanwong: Actually, if I try this with a large ts with no node metadata (e.g. `ts = msprime.simulate(10000, recombination_rate=10000)`), then `_change_ragged_order2` takes about 4 times longer than my `_change_ragged_order` original version. So 🤷

hyanwong: Actually, this would seem to hit the common use-cases:

```python
def change_ragged_order(ragged_arr, ragged_offset, new_order):
    # returns a tuple of a new ragged_col in a new order and a new offset array
    if len(ragged_arr) == 0:
        # Common edge case: no metadata => offset array is just zeros
        return ragged_arr, np.zeros(len(new_order), dtype=ragged_offset.dtype)
    tskit.unpack_arrays(ragged_arr, ragged_offset)
    return tskit.pack_arrays([arr[i] for i in new_order])

```

hyanwong: Changed now.

hyanwong:
After a quick chat with Jerome, we think that (a) this routine is mostly tsinfer-specific so (b) we should probably always try to set the metadata, and simply warn if this is not possible. This could probably apply to the other tsinfer-set metadata fields like ancestor_data_id too.

hyanwong:
OK, I think this is ready to go. Does it look OK to you @nspope ?

tests/test_util.py
```diff
@@ -0,0 +1,243 @@
...
+import msprime
+import numpy as np
+import pytest
+import tskit
+
+import tsdate
+
+
+class TestSplitDisjointNodes:
+    def test_nosplit(self):
```
nspope, line 37: would also be good to test that split_disjoint_nodes leaves a tree sequence unchanged if it's already had the routine called on it

nspope: also it looks like I stuck "TestNodeSplitting" class into test_functions.py, let's move it over to this file

nspope (approved):
LGTM, just a couple comments re: tests. Also, just out of curiosity, I'm wondering what the timing of this routine on the 40K GEL data is, versus the old version (that didn't do anything with metadata). If there's a sizeable gap, lets open an issue about moving some of the internals into numba.

tests/test_util.py
hyanwong: Great, done. I'll get Ben to test the speed, then merge.

hyanwong:
I didn't test on the 40K ts, but I did on chr2 of the unified genealogy (7524 samples), which isn't too bad a test, as `split_disjoint_nodes` still creates half-a-million new nodes. For reference, it takes about 75 seconds to simplify this ts (i.e. removing unary nodes) before splitting.

The old routine took 15 secs to run. The new one, having set a permissive_json schema, takes 25 secs. I think that's fine, so I'll merge.

At the moment, tsinfer doesn't set a node metadata schema, so the metadata is not saved anyway (and so it only takes 15 secs even with the new code). When we fix _tsinfer_ to set a proper schema, it will take a little longer on large tree sequences, but still a fraction of the time required for simplification.

PR #383, opened 2024-06-04
Set a new descriptive schema on mutations if none exists

hyanwong:
And emit warning if mutation metadata cannot be set. This also means that existing mutation metadata is kept. However, since we now validate the metadata, it increases the time taken. E.g. dating the unified genealogy chr 2q the metadata set goes from 84 seconds to 403 seconds on my laptop (for comparison, the 10 EP steps take ~ 7 mins).

There might be a way to do this faster without validation if no node metadata already exists, I suppose.


hyanwong:
New code avoids decoding the mutation metadata if there is none there. This reduces the time to 310 seconds for the metadata setting on chr2q. I think this is probably a fair compromise: it's always going to take time to validate 12 million JSON entries rather than just dumping raw bytes.

hyanwong:
Aha - it's the validation that takes time: the encoding is quick. I've made a change to only bother validating if there is existing data, and that brings it down to 95 secs

hyanwong:
I'm just going to merge this now. Is that OK @nspope ? Do you want to look over it beforehand?

Issue #376, opened 2024-05-13
Move population_size and priors out of .date() method


hyanwong:
Now that the "variational_gamma" method doesn't need to be passed a "population_size" or "priors" object, we should probably move those parameters into the `inside_outside` and `maximization` methods, rather than being in the top-level `.date()` function.

hyanwong:
done

PR #385, opened 2024-06-04
Document (and undocument) standard params & tidy calling conventions

hyanwong:


hyanwong:
Fixes https://github.com/tskit-dev/tsdate/issues/376

PR #405, opened 2024-06-07
Fix changelog

benjeffery:
@hyanwong I was confused by the changelog here - is this correct?


hyanwong:
Ah, I had forgotten about 0.1.7, sorry! Looking back on it, yes, what you have in the PR is right.

I think the new release should be 0.2, not 0.1.8 though. The change to default vgamma is substantial.

PR #416, opened 2024-07-10
Update msprime in docs

benjeffery:


hyanwong:
LGTM, but is failing with the "numpy.core.multiarray failed to import" error. I assume this is a numpy 2.0 thing. Presumably we need to update tsinfer to allow numpy2.0 first, then fix tsdate?

benjeffery:
Weird, even specifying no-binary for tsinfer still fails. I need to get a tsinfer release out anyway.

hyanwong:
I think this was fixed in #417. I unpinned msprime in the docs requirements. Is there any reason to pin to specific versions for docs? Sometimes it's helpful to take advantage of newer released features.

I.e., can I just close this PR without merging?

jeromekelleher:
I think the idea of pinning for docs is to make things less likely to break over time. So, you have a stable set of dependencies, and aren't fixing the build all the time. I think this is probably the right approach - things have just broken quite a lot recently because we've neglected the ecosystem for a while, coupled with the major change from numpy 2.

hyanwong:
I'm happy to add pinning back in. I note that some libraries are pinned in `docs/requirements.txt` and some aren't: should we either pin all or none (except, I guess, sphinx-book-theme, as per comments in the file).

Probably we can remove some of the dependencies too.

PR #404, opened 2024-06-06
Switch to ruff linting

benjeffery:

PR #421, opened 2024-07-24
Switch to ruff linting

hyanwong:
Replaces #404


hyanwong:
I'm just going to merge this so that we don't get stuck having to re-lint new code. These should all be formatting changes

PR #422, opened 2024-07-24
Use newer numpy recarray calling convention to avoid warnings

hyanwong:


hyanwong:
Actually now the same function

PR #426, opened 2024-07-27
Fix references to "flat" priors in docs

nspope:
as we actually do use a "prior" (really, regularisation) on root node ages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@
   mutations via `singletons_phased=False` option. The API is preliminary and
   may change.
 
+**Documentation**
+
+- Fixed description of priors for variational gamma method, which were referred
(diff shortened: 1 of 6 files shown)


hyanwong:
But we do use a flat prior on non-root ages, right?

hyanwong:
I think it would be good to keep the idea that the prior on non-roots is essentially flat (after all, these are most of the nodes we are trying to date: the roots are an exception)

nspope:
What we're doing is using regularisation on the roots-- fit by EM. If viewed as a prior (ignoring the EM) this induces a distribution on the internal nodes, too. I agree, it's much less strong than using informative priors on every node. But I think it's good to be clear about what we're doing here, until we write up the algorithm.

hyanwong (approved):

PR #433, opened 2024-11-01
Allow internal/non-contemporary samples for dating sc2ts, pedigrees

nspope:
A minimal case:
```python
Mutations: 16

--- TRUE TREES ---
13.00┊               ┊             ┊             ┊       16    ┊
     ┊               ┊             ┊             ┊     ┏━━┻━━┓ ┊
12.00┊  15           ┊      15     ┊             ┊     ┃     ┃ ┊
     ┊ ┏━┻━━┓        ┊   ┏━━━┻━━┓  ┊             ┊     ┃     ┃ ┊
8.00 ┊ ┃    ┃        ┊  13      ┃  ┊    13       ┊    14     ┃ ┊
     ┊ ┃    ┃        ┊ ┏━┻━┓    ┃  ┊ ┏━━┳┻━━━┓   ┊  ┏━━┻━┓   ┃ ┊
7.00 ┊ ┃   11        ┊ ┃   ┃   11  ┊ ┃ 11    ┃   ┊ 12    ┃   ┃ ┊
     ┊ ┃ ┏━┳┻━━┓     ┊ ┃   ┃   ┏┻┓ ┊ ┃ ┏┻┓   ┃   ┊ ┏┻┓   ┃   ┃ ┊
6.00 ┊ 1 2 3   0     ┊ 2   1   3 0 ┊ 2 3 0   1   ┊ 2 0   1   3 ┊
     ┊         ┃     ┊     ┃     ┃ ┊     ┃   ┃   ┊   ┃   ┃     ┊
3.00 ┊        10     ┊     ┃     ┃ ┊     ┃   ┃   ┊   ┃   ┃     ┊
     ┊       ┏━┻━┓   ┊     ┃     ┃ ┊     ┃   ┃   ┊   ┃   ┃     ┊
2.00 ┊       ┃   ┃   ┊     ┃     ┃ ┊     ┃   9   ┊   ┃   9     ┊
     ┊       ┃   ┃   ┊     ┃     ┃ ┊     ┃ ┏━┻┓  ┊   ┃  ┏┻━┓   ┊
1.00 ┊       ┃   8   ┊     8     ┃ ┊     ┃ ┃  8  ┊   ┃  8  ┃   ┊
     ┊       ┃ ┏━╋━┓ ┊   ┏━╋━┓   ┃ ┊     ┃ ┃ ┏┻┓ ┊   ┃ ┏┻┓ ┃   ┊
0.00 ┊       4 5 6 7 ┊   5 6 7   4 ┊     4 5 6 7 ┊   4 5 7 6   ┊
     0               1             2             3             4

--- DATED WITH INTERNAL SAMPLES ---
15.85┊               ┊             ┊             ┊       16    ┊
     ┊               ┊             ┊             ┊     ┏━━┻━━┓ ┊
15.47┊  15           ┊      15     ┊             ┊     ┃     ┃ ┊
     ┊ ┏━┻━━┓        ┊   ┏━━━┻━━┓  ┊             ┊     ┃     ┃ ┊
10.35┊ ┃    ┃        ┊  13      ┃  ┊    13       ┊     ┃     ┃ ┊
     ┊ ┃    ┃        ┊ ┏━┻━┓    ┃  ┊ ┏━━┳┻━━━┓   ┊     ┃     ┃ ┊
9.33 ┊ ┃    ┃        ┊ ┃   ┃    ┃  ┊ ┃  ┃    ┃   ┊    14     ┃ ┊
     ┊ ┃    ┃        ┊ ┃   ┃    ┃  ┊ ┃  ┃    ┃   ┊  ┏━━┻━┓   ┃ ┊
8.19 ┊ ┃   11        ┊ ┃   ┃   11  ┊ ┃ 11    ┃   ┊  ┃    ┃   ┃ ┊
     ┊ ┃ ┏━┳┻━━┓     ┊ ┃   ┃   ┏┻┓ ┊ ┃ ┏┻┓   ┃   ┊  ┃    ┃   ┃ ┊
6.95 ┊ ┃ ┃ ┃   ┃     ┊ ┃   ┃   ┃ ┃ ┊ ┃ ┃ ┃   ┃   ┊ 12    ┃   ┃ ┊
     ┊ ┃ ┃ ┃   ┃     ┊ ┃   ┃   ┃ ┃ ┊ ┃ ┃ ┃   ┃   ┊ ┏┻┓   ┃   ┃ ┊
6.00 ┊ 1 2 3   0     ┊ 2   1   3 0 ┊ 2 3 0   1   ┊ 2 0   1   3 ┊
     ┊         ┃     ┊     ┃     ┃ ┊     ┃   ┃   ┊   ┃   ┃     ┊
2.72 ┊        10     ┊     ┃     ┃ ┊     ┃   ┃   ┊   ┃   ┃     ┊
     ┊       ┏━┻━┓   ┊     ┃     ┃ ┊     ┃   ┃   ┊   ┃   ┃     ┊
2.11 ┊       ┃   ┃   ┊     ┃     ┃ ┊     ┃   9   ┊   ┃   9     ┊
     ┊       ┃   ┃   ┊     ┃     ┃ ┊     ┃ ┏━┻┓  ┊   ┃  ┏┻━┓   ┊
0.98 ┊       ┃   8   ┊     8     ┃ ┊     ┃ ┃  8  ┊   ┃  8  ┃   ┊
     ┊       ┃ ┏━╋━┓ ┊   ┏━╋━┓   ┃ ┊     ┃ ┃ ┏┻┓ ┊   ┃ ┏┻┓ ┃   ┊
0.00 ┊       4 5 6 7 ┊   5 6 7   4 ┊     4 5 6 7 ┊   4 5 7 6   ┊
     0               1             2             3             4
```
<details>

<summary>Code to simulate and run on test case</summary>

```python
import msprime
import numpy as np
import tskit


def simulate_internal_samples(pop_size=2, sample_time=6, L=4, r=0.1, random_seed=1):
    """ 
    Simulate ts with noncontemporary samples that have descendants 
    https://github.com/tskit-dev/msprime/discussions/2260
    """
    rng = np.random.default_rng(random_seed)
    
    samples = [
        msprime.SampleSet(pop_size),
        msprime.SampleSet(1, ploidy=1, time=sample_time + 1),
    ]
    
    # recent history up to sample time
    ts1 = msprime.sim_ancestry(
        samples=samples,
        population_size=pop_size,
        model="dtwf",
        end_time=sample_time,
        sequence_length=L,
        recombination_rate=r,
        random_seed=random_seed,
    )
    
    # remove dummy node above the sample time
    ts1 = ts1.simplify(range(2 * pop_size), keep_unary=True)
    
    # history before sample time
    ts2 = msprime.sim_ancestry(
        samples=[msprime.SampleSet(pop_size, time=sample_time)],
        model="dtwf",
        population_size=pop_size,
        sequence_length=L,
        recombination_rate=r,
        random_seed=random_seed + 1000,
    )
    
    # remap roots to samples in ts2
    roots = [n.id for n in ts1.nodes() if n.time == sample_time]
    tips = ts2.samples()
    rng.shuffle(tips)
    node_mapping = [tskit.NULL for _ in ts1.nodes()]
    for t, n in zip(tips[:len(roots)], roots):
        node_mapping[n] = t
    ts = ts2.union(ts1, node_mapping, check_shared_equality=False)

    return ts


# ---- some minimal test cases --- #

import tsdate

def get_internal_samples(ts):
    always_internal = np.full(ts.num_nodes, False)
    always_internal[list(ts.samples())] = True
    for t in ts.trees():
        for n in t.samples():
            if t.num_children(n) == 0:
                always_internal[n] = False
    return np.flatnonzero(always_internal)

mu = 0.1
ts = simulate_internal_samples(random_seed=1).simplify()
ts = msprime.sim_mutations(ts, rate=mu, random_seed=1)
print(f"Mutations: {ts.num_mutations}")
print(f"--- TRUE TREES ---\n", ts.draw_text())

# (1) case with internal-only samples (may be unary) and noncontemporary leaf samples
# NB: sample ages can be modified, because of how we force positive branch lengths.
#     this'll be fixed (or a least minimized) by setting `constr_iterations`
foo = tsdate.date(ts, mutation_rate=mu, rescaling_intervals=1, constr_iterations=100)
assert np.allclose(foo.nodes_time[list(foo.samples())], ts.nodes_time[list(foo.samples())])
print(f"--- DATED WITH INTERNAL SAMPLES ---\n", foo.draw_text())

# (2) when there's a unary node that's not a sample, we should fail
internal_samples = get_internal_samples(ts)
tab = ts.dump_tables()
new_flags = tab.nodes.flags.copy()
new_flags[internal_samples] = 0
tab.nodes.flags = new_flags
ts_no_internal = tab.tree_sequence()
try:
    bar = tsdate.date(ts_no_internal, mutation_rate=mu, rescaling_intervals=1, constr_iterations=100)
except ValueError:
    pass

# get rid of unary node, and it'll work
bar = tsdate.date(ts_no_internal.simplify(), mutation_rate=mu, rescaling_intervals=1, constr_iterations=100)
```

</details>

Some caveats:
- Internal samples can be unary over part or all of their span; but we'll still error out if non-sample nodes are unary
- Because of how we're forcing the positive branch length constraint, ages of internal samples (e.g. with descendants) can be pushed upward. This can be fixed (or minimized) by setting `constr_iterations` to 100 or something similar.
- Rescaling is done as usual, but without changing the ages of sample nodes. This might cause issues with nodes "bunching up" if e.g. the mutation rate is misspecified or the sample age is wrong or there's other weirdness with the inputs. There's better ways to do this, but it'll take a bit more work to get something that's reliable.
- I've only tried this out on minimal examples, and not inferred TS's, as I'm not quite clear on how to add internal samples in this case (maybe it's straightforward). @hyanwong maybe you could try this out on sc2ts?

### DIFF
--- a/tests/test_noncontemporary.py
+++ b/tests/test_noncontemporary.py
@@ -1,0 +1,176 @@
+# MIT License
+#
+# Copyright (c) 2021-25 Tskit Developers
+# Copyright (c) 2020-21 University of Oxford
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Test cases for dating noncontemporary samples
+"""
+
+import msprime
+import numpy as np
+import pytest
+import tskit
+
+from tsdate import date
+
+
+def simulate_internal_samples(pop_size=2, sample_time=6, L=4, r=0.1, random_seed=1):
+    """
+    Simulate ts with noncontemporary samples that have descendants
+    https://github.com/tskit-dev/msprime/discussions/2260
+    """
+    rng = np.random.default_rng(random_seed)
+
+    samples = [
+        msprime.SampleSet(pop_size),
+        msprime.SampleSet(1, ploidy=1, time=sample_time + 1),
+    ]
+
+    # recent history up to sample time
+    ts1 = msprime.sim_ancestry(
+        samples=samples,
+        population_size=pop_size,
+        model="dtwf",
+        end_time=sample_time,
+        sequence_length=L,
+        recombination_rate=r,
+        random_seed=random_seed,
+    )
+
+    # remove dummy node above the sample time
+    ts1 = ts1.simplify(range(2 * pop_size), keep_unary=True)
+
+    # history before sample time
+    ts2 = msprime.sim_ancestry(
+        samples=[msprime.SampleSet(pop_size, time=sample_time)],
+        model="dtwf",
+        population_size=pop_size,
+        sequence_length=L,
+        recombination_rate=r,
+        random_seed=random_seed + 1000,
+    )
+
+    # remap roots to samples in ts2
+    roots = [n.id for n in ts1.nodes() if n.time == sample_time]
+    tips = ts2.samples()
+    rng.shuffle(tips)
+    node_mapping = [tskit.NULL for _ in ts1.nodes()]
+    for t, n in zip(tips[: len(roots)], roots):
+        node_mapping[n] = t
+    ts = ts2.union(ts1, node_mapping, check_shared_equality=False)
+
+    return ts
+
+
+def get_internal_samples(ts):
+    always_internal = np.full(ts.num_nodes, False)
+    always_internal[list(ts.samples())] = True
+    for t in ts.trees():
+        for n in t.samples():
+            if t.num_children(n) == 0:
+                always_internal[n] = False
+    return np.flatnonzero(always_internal)
+
+
+class TestInternalConstraints:
+    """
+    Test dating when some internal nodes are fixed
+
+    NB: Currently we need to use `constr_iterations` to
+    ensure that the constraint forcing doesn't modify sample
+    ages
+    """
+
+    def test_with_unary_internal_samples(self):
+        mu = 0.1
+        ts = simulate_internal_samples(random_seed=1).simplify()
+        ts = msprime.sim_mutations(ts, rate=mu, random_seed=1)
+        dts = date(ts, mutation_rate=mu, rescaling_intervals=1)
+        assert np.allclose(
+            dts.nodes_time[list(dts.samples())],
+            ts.nodes_time[list(ts.samples())],
+        )
+
+    def test_with_nonunary_internal_samples(self):
+        rng = np.random.default_rng(1)
+        ts = msprime.sim_ancestry(
+            samples=10,
+            population_size=1e4,
+            sequence_length=1e5,
+            recombination_rate=1e-8,
+            random_seed=1,
+        )
+        ts = msprime.sim_mutations(ts, rate=1e-8, random_seed=1)
+        fixed_nodes = np.unique(rng.integers(ts.num_samples, ts.num_nodes, size=10))
+        nodes_flags = ts.nodes_flags.copy()
+        nodes_flags[fixed_nodes] = tskit.NODE_IS_SAMPLE
+        tab = ts.dump_tables()
+        tab.nodes.flags = nodes_flags
+        tab.sort()
+        ts = tab.tree_sequence()
+        dts = date(ts, mutation_rate=1e-8, rescaling_intervals=1)
+        assert np.allclose(
+            dts.nodes_time[fixed_nodes],
+            ts.nodes_time[fixed_nodes],
+        )
+
+    def test_that_unary_nonsample_fails(self):
+        mu = 0.1
+        ts = simulate_internal_samples(random_seed=1).simplify()
+        ts = msprime.sim_mutations(ts, rate=mu, random_seed=1)
+        internal_samples = get_internal_samples(ts)
+        tab = ts.dump_tables()
+        new_flags = tab.nodes.flags.copy()
+        new_flags[internal_samples] = 0
+        tab.nodes.flags = new_flags
+        ts_no_internal = tab.tree_sequence()
+        with pytest.raises(ValueError):
+            date(
+                ts_no_internal,
+                mutation_rate=mu,
+            )
+
+
+class TestUnconstrainedSamples:
+    """
+    Test dating when some leaf nodes do not have a fixed time
+    """
+
+    def test_with_free_leaf_nodes(self):
+        ts = msprime.sim_ancestry(
+            samples=10,
+            population_size=1e4,
+            sequence_length=1e5,
+            recombination_rate=1e-8,
+            random_seed=1,
+        )
+        ts = msprime.sim_mutations(ts, rate=1e-8, random_seed=1)
+        free_nodes = np.array([1, 3, 18]).astype(np.int32)
+        nodes_flags = ts.nodes_flags.copy()
+        nodes_flags[free_nodes] = 0
+        tab = ts.dump_tables()
+        tab.nodes.flags = nodes_flags
+        tab.sort()
+        ts = tab.tree_sequence()
+        dts = date(ts, mutation_rate=1e-8, rescaling_intervals=1)
+        assert np.all(ts.nodes_time[free_nodes] == 0.0)
+        assert np.all(dts.nodes_time[free_nodes] > 0)
--- a/tsdate/approx.py
+++ b/tsdate/approx.py
@@ -140,18 +140,22 @@ def approximate_gamma_mom(mean, variance):
     return shape - 1.0, rate
 
 
-@numba.njit(_unituple(_f, 2)(_f, _f, _f, _f))
-def approximate_gamma_iqr(q1, q2, x1, x2):
+@numba.njit(_unituple(_f, 2)(_f, _f, _f, _f, _f))
+def approximate_gamma_iqr(q1, q2, x1, x2, max_shape):
     """Find gamma natural parameters that match empirical quantiles"""
+
+    def upper_bound(q, x):  # cap shape parameter
+        beta = hypergeo._gammainc_inv(max_shape, q) / x
+        return max_shape - 1, beta
+
+    if x2 == x1:
+        return upper_bound(q1, x1)
     if not (q2 > q1 and x2 > x1):
         raise KLMinimizationFailedError("Quantiles must be sorted")
-    # find starting value from asymptotic solutions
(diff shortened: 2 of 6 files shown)


hyanwong:
This is great!

I'm just thinking about how I might get a sensible mutation rate to use. I presume one way is to ensure that the TS is simplified, then count up the total mutational area and divide by the number of mutations, i.e.

```
rate = ts.num_mutations / ((ts.edges_right - ts.edges_left) * (ts.nodes_time[ts.edges_parent] - ts.nodes_time[ts.edges_child])).sum()
```

hyanwong:
An initial run on a huge arg gives:

```
File [~/Library/jupyterlab-desktop/jlab_server/lib/python3.12/site-packages/tsdate/variational.py:693](http://localhost:59486/lab/tree/~/Library/jupyterlab-desktop/jlab_server/lib/python3.12/site-packages/tsdate/variational.py#line=692), in ExpectationPropagation.iterate(self, max_shape, min_step, em_maxitt, em_reltol, regularise, check_valid)
    691 # exponential regularization on roots
    692 if regularise:
--> 693     self.propagate_prior(
    694         self.unconstrained_roots,
    695         self.node_posterior,
    696         self.node_factors,
    697         self.node_scale,
    698         max_shape,
    699         em_maxitt,
    700         em_reltol,
    701     )
    703 # absorb the scaling term into the factors
    704 self.rescale_factors(
    705     self.edge_parents,
    706     self.edge_children,
   (...)
    711     self.node_scale,
    712 )

ZeroDivisionError: division by zero

```

I assume there is some deeper number stuff going on: is there a way to switch number off to get a deeper trace?

hyanwong:
Probably the reason for this is that there are very few non-sample nodes. For example, if I simplify to the first 50 nodes, I get a tree sequence consisting of a single tree like this
Screenshot 2024-11-01 at 22 21 26

[small_ts.trees.zip](https://github.com/user-attachments/files/17604594/small_ts.trees.zip)

hyanwong:
Here's a very simple case which I think *should* work, but gives a division-by-zero error:

Screenshot 2024-11-01 at 22 31 37

[tiny_ts.trees.zip](https://github.com/user-attachments/files/17604625/tiny_ts.trees.zip)

nspope:
In several places in the code, I'm assuming the time scale starts at zero, but it seems this is never checked. I suppose we should enforce this. Can you try it with (a) shifting nodes time so that the most recent sample node is at 0, (b) turning rescaling off with rescaling_intervals=0?

hyanwong:
Good point. I have been using rescaling_intervals=0 for this. When I transform the times to start at 0, and flag the root(s) as not-a-sample, I can get dating to work with pretty large subsets of the data (e.g. 350k samples), which is fantastic.

It still fails on larger sample sizes though, but now with a slightly different (assertion) error:

```
File [~/Library/jupyterlab-desktop/jlab_server/lib/python3.12/site-packages/tsdate/variational.py:89](http://localhost:59486/lab/tree/~/Library/jupyterlab-desktop/jlab_server/lib/python3.12/site-packages/tsdate/variational.py#line=88), in _damp()
     87 b = 1.0 if (x[1] - y[1] > x[1] * s) else (1 - s) * x[1] [/](http://localhost:59486/) y[1]
     88 d = min(a, b)
---> 89 assert 0.0 < d <= 1.0
     90 return d

AssertionError:
```

nspope:
This is probably due to floating point error, in that comparison against 1. Seems like like this a good case to catch these sorts of numerical issues. Can you share the trees over slack or email, and the code you're using to prepare them?

jeromekelleher:
Amazing!

hyanwong:
Just to post some of @nspope's thoughts. The most recent branch fixes the sample-root issue and also allows rescaling, but it's not clear to Nate or myself that rescaling is sensible here anyway. Re the `0.0 < d <= 1.0` assertion fail he says:
> It does seem like we're hitting against similar issues with these trees as we were when trying to date TS's with unary nodes -- there's lots of constraints, so the gamma shape parameters go to the limit. I'd suggest comparing dates with different numbers of EP iterations to see if convergence is really slow. If it is, then rescaling might help.

We can use the current code to date relatively large portions of the sc2ts ARGs, so we could try cho9pping them into chunks to get something working. Alternatively, I wonder if there is a way to skip certain edges if they create particularly problematic constraints, and output the list of skipped edges separately? This could be useful to identify badly inferred relationships. For this I suggest that it would be much better to omit bad edges than bad nodes, if possible.

nspope:
I'd skip rescaling for the covid trees -- there's so many constraints (including a fixed root age) that I really doubt that it's necessary (in unconstrained ARGs it's generally most helpful towards the root).

Regarding the assertion -- I can do something hacky to bypass the issue, if you want to give it a go on the full ARG this or next week (and do it properly later).

nspope:
With last commit it'll run through on the 450K covid example you sent @hyanwong -- can you try it on a larger example?

hyanwong:
At the end of the dating process, we need to pick dates for the nodes (and mutations) that are consistent with the parent-older-than-child requirement of a tree sequence. There are two parts to this, a least squares and a simple upwards iteration. However both of these can cause samples at non-zero times to have their times changed, which is not ideal.

A relatively simple fix would be to rescale all the node times at the end of the routine. If we have a fixed node N2 at time t2, closest in time to a younger fixed node N1 at times t1, but we estimate its time as t2*, we simply need to multiply the times of non-fixed nodes at times between t1 and t2* by (t2-t1)/(t2*-t1).

hyanwong:
Here's an (untested) algorithm for reselling using a single pass through the nodes in time order. It requires use to ensure that all sample nodes that occur at the same time are given the same inferred time, which we should be able to to using the iterative approach, but which is not coded up below.

```
# TODO: use the iterative approch to ensure that all sample nodes that occur at
# the same time are given the same inferred time, and in the same order

tmp_sample_times = np.unique(tmp_ts.nodes_time[node_is_sample])

if len(tmp_sample_times) > 1:
    orig_sample_times = orig_ts.nodes_time[node_is_sample]
    tmp_times = tmp_ts.nodes_time
    new_times = np.zeros_like(ts.nodes_time)
    idx = 0
    rescaling = None
    for u in np.lexsort((np.arange(tmp_ts.num_nodes), tmp_ts.nodes_time)):
        time_u = tmp_times[u]
        if rescaling is None and not node_is_sample[u]:
            pass  # Skip until the youngest sample node is found
        elif time_u == tmp_sample_times[idx]:
            # Same time as the sample
            new_times[u] = orig_sample_times[idx]
        else:
            tmp_diff = time_u - tmp_sample_times[idx]
            if idx == len(orig_sample_times) - 1:
                # top of the ARG, just subtract a constant
                new_times[u] = orig_sample_times[idx] + tmp_diff
            else:
                while idx < len(orig_sample_times) - 1 and time_u >= tmp_sample_times[idx + 1]:
                    # new timeslice
                    idx += 1
                    tmp_diff = time_u - tmp_sample_times[idx]
                    try:
                        rescaling = (orig_sample_times[idx + 1] - orig_sample_times[idx]) / (tmp_sample_times[idx + 1] - tmp_sample_times[idx])
                    except:
                        raise
                new_times[u] = orig_sample_times[idx] + tmp_diff * rescaling
        
tables = tmp_ts.dump_tables()
tables.nodes.time = new_times
new_ts = tables.tree_sequence()
```

hyanwong:
I'd like to merge this @nspope . Is there any reason not to? For example, I see `# FIXME: this is a hacky bypass` in one place (but I think that's fine, even if we release this code as a new version, right)

hyanwong:
@mergifyio rebase

nspope:
Don't merge it, because it's doing the wrong thing (resetting messages in one part of the algorithm). That'll be fine for sc2ts where there's no prior (roots are fixed) and no unphased singletons. Otherwise it'll do the wrong thing.

hyanwong:
Ah, I hadn't realised that (or had forgotten). Thanks for the explanation.

nspope:
Hey @hyanwong -- I think this is ready, could you please run on sc2ts and see that you're getting the same dates as earlier?

nspope:
No rush, enjoy your vacation!

hyanwong:
I just tried this out and the new nonsample dates are within 1e-6 of the previous version, apart from the root, which is ~0.6 different (was 1192.6696, now 1191.9805).

This looks OK to me. Shall I merge it?

nspope:
Great! I should check that the "typical" use case (no internal constraints) gives identical answers before/after this PR, as I had to mess with the internals a little. I'll do that later today; then we can merge.

nspope:
Sorry for the delay here-- this is good to merge in, I get identical results compared to main when there's no internal constraints.

PR #441, opened 2024-12-01
Allow the numba cache to be used, for development

### LINKED ISSUE #438: Reduce import time
The import time (due to numba AOT compilation) is driving me nuts-- I wonder if there's a way we can build an extension optionally on install (e.g. triggered by a flag to pip). Jerome and others have had bad experiences with numba extensions, but 30 to 60 seconds for import really slows down the debug cycle, and if extension-building is turned off by default I don't think it'd be an issue.

hyanwong:
Copied from SGKIT. Fixes #438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,12 @@
 **Features**
 
 - An `allow_unary` flag (``False by default``) has been added to all methods.
+
 - A `set_metadata` flag has been added so that node and mutation metadata can be
   omitted, saved (default), or overwritten even if this requires changing the schema.
+
+- An environment variable `TSDATE_ENABLE_NUMBA_CACHE` can be set to cache JIT
+  compiled code, speeding up loading time (useful when testing).
 
 **Documentation**
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -35,3 +35,24 @@ or
 
 Alternatively, the {ref}`Python API <sec_python_api>` allows more fine-grained control
 of the inference process.
+
+(sec_installation_testing)=
+
+## Testing
+
+Unit tests can be run from a clone of the
+[Github repository](https://github.com/tskit-dev/tsdate) by running pytest
+at the top level of the repository
+
+    $python -m pytest
+
+_Tsdate_ makes extensive use of [numba](https://numba.pydata.org)'s
+"just in time" (jit) compilation to speed up time-consuming numerical functions.
+Because of the need to compile these functions, loading the tsdate package can take
+tens of seconds. To speed up loading time, you can set the environment variable
+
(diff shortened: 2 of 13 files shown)


nspope:
I ran into the same issues as you here-- the bulk of stuff is still being compiled on import. I think pulling all the EP stuff into a low-level extension is the way to go.

hyanwong:
> I think pulling all the EP stuff into a low-level extension is the way to go.

Can we do this while keeping it in python? Or are you thinking of writing the low-level e.g. in rust?

nspope:
It may be possible to build the extension with numba, but I think it'd be stabler and easier to maintain a C or rust extension.

nspope:
I could make the edits to the scipy special functions on this branch if you'd like @hyanwong ?

hyanwong:
Yes, why not. Thank you.

nspope:
Sure, sorry to drop the ball on this. Where I'm thinking is, https://github.com/tskit-dev/tsdate/blob/240593d01a1a77561dadebee82cfb5fae67f83bc/tsdate/hypergeo.py#L72C17-L74C26

the _gammaln_f8 should be changed to math.lgamma -- hopefully that's enough to make more functions cache-able?

hyanwong:
No problem,. I'll give it a whirl. Thanks for the pointer.

hyanwong:
OK, this seems to work, and cuts import down from 29 seconds to 9 seconds on my laptop.

I put `_hyp2f1_unity` as an inner function within `_hyp2f1_laplace`, as I couldn't figure out how to pass things neatly otherwise. I guess you could rework that if it grates or is expected to make things slow @nspope. I haven't figured out how to make `_gammainc_inv` cacheable either, so that would make everything faster too.

hyanwong:
Aha. I spotted some more functions I could mark to cache. Now down to 5 secs for import.

hyanwong:
I guess this is good to go, at least for the time being.

To enable the caching (e.g. for development), you need to explicitly set `TSDATE_DISABLE_NUMBA_CACHE=0` which seems a bit of a weird way to phrase it (I would have chosen `TSDATE_ENABLE_NUMBA_CACHE=1` I think). I just copied that from sgkit: is there any reason for the choice of name / polarisation in sgkit, @jeromekelleher ?

jeromekelleher:
I don't think there's a good reason for the choice of name, I agree with your preference for ``TSDATE_ENABLE_NUMBA_CACHE=1``

hyanwong:
@nspope - I'm happy to merge this. Loading time for the library is now down to 2.5s on my laptop with `TSDATE_ENABLE_NUMBA_CACHE=1`. Does it look OK to you?

nspope (approved):
Awesome, thanks @hyanwong !

PR #442, opened 2024-12-01
Record resources used in tsdate inference

hyanwong:
This should make it easier to create perf plots, etc.

The max_memory usage requirements may not be very reliable, when e.g. the `date()` method is called within a single script.


hyanwong:
I'm just going to merge this, as it seems to work fine, and is pretty minor

PR #448, opened 2024-12-13
Change the default number of ep_iterations to 25

hyanwong:
Fixes 447. I also made sure we weren't specifying defaults in 2 different places. I think this is OK, right, @nspope ?


nspope (approved):
LGTM-- I do like having defaults set for EP as I run it standalone, but agree that it's sloppy to have them defined multiple places.

PR #449, opened 2024-12-14
Add `set_metadata` parameter

### LINKED ISSUE #302: "set_metadata" parameter

hyanwong:
Fixes #302. As described in that issue, we issue a warning and don't write metadata for old tsinfer versions (these have no schema, and raw byte metadata in the nodes table)

### DIFF
--- a/tsdate/core.py
+++ b/tsdate/core.py
@@ -87,11 +87,12 @@ class EstimationMethod:
         allow_unary=None,
         record_provenance=None,
         constr_iterations=None,
+        set_metadata=None,
         progress=None,
         # Deprecated params
         return_posteriors=None,
     ):
-        # Set up all the generic params describe in the tsdate.date function, and define
+        # Set up all the generic params described in the tsdate.date function, and define
         # priors if not passed-in already
         if return_posteriors is not None:
             raise ValueError(
@@ -106,6 +107,7 @@ class EstimationMethod:
         self.recombination_rate = recombination_rate
         self.return_fit = return_fit
         self.return_likelihood = return_likelihood
+        self.set_metadata = set_metadata
         self.pbar = progress
         self.time_units = "generations" if time_units is None else time_units
         if record_provenance is None:
@@ -206,7 +208,7 @@ class EstimationMethod:
         # Add posterior mean and variance to node/mutation metadata
         meta_timing = time.time()
         self.set_time_metadata(
-            nodes, node_mean_t, node_var_t, schemas.default_node_schema, overwrite=True
+            nodes, node_mean_t, node_var_t, schemas.default_node_schema
         )
         self.set_time_metadata(
             mutations, mut_mean_t, mut_var_t, schemas.default_mutation_schema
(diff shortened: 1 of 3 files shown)


tsdate/core.py
```diff
@@ -226,40 +228,43 @@ def get_modified_ts(self, result, eps):
...
-                    return
+    def set_time_metadata(self, table, mean, var, default_schema):
+        # Try to set metadata: if we fail, clear metadata, reset schema, and try again
+        def _time_md_array(table, mean, var):
+            # Return an array of metadata dicts, or raise an error if
+            # schema is None or metadata is not valid
+            s = table.metadata_schema
+            if s.schema is None:
+                raise tskit.MetadataEncodingError("No schema set")
+            encoder = s.encode_row if s == default_schema else s.validate_and_encode_row
```
benjeffery, line 239: I'm not quite getting why we have to have the `if` here? Why does default schema not have `validate_and_encode_row`?

tsdate/core.py
```diff
@@ -226,40 +228,43 @@ def get_modified_ts(self, result, eps):
...
+        if self.set_metadata is False or var is None:
+            return  # no md to set (e.g. outside maximization method)
+        assert len(mean) == len(var) == table.num_rows
+        try:
+            table.packset_metadata(_time_md_array(table, mean, var))
+        except (tskit.MetadataEncodingError, tskit.MetadataValidationError) as e:
+            table_name = type(table).__name__
+            if len(table.metadata) > 0 or table.metadata_schema.schema is not None:
+                if not self.set_metadata:
+                    logger.warning(f"Could not set time metadata on {table_name}: {e}")
```
benjeffery, line 260: Could add a pointer here to the override flag that would suppress the warning?

benjeffery (commented):
Seems to make sense, couple of comments. I assume the other tests test the `None` default case?

tsdate/core.py
hyanwong: Oh, good point. I put that in because I didn't want the hassle of validating json metadata. but actually this would probably not cause efficiency issues, so I guess we can simply call `validate_and_encode_row` for json data too?

benjeffery: Yeah, I'd rather have a single code path here unless and until we have a proven bottleneck.

PR #458, opened 2025-03-29
Remove tscompare routines

### LINKED ISSUE #311: Move CladeMap and associated functions into tscompare repo
See https://github.com/tskit-dev/tsdate/pull/310#issuecomment-1691210197

hyanwong:
I think all these evaluation.py routines are now present and tested (using the same code) in tscompare, right @nspope ? I couldn't find where they were used in this repo anyway. Does this PR removing them make sense?

Fixes #311


nspope:
Yup, these can be removed from tsdate.

PR #485, opened 2026-01-22
Disambiguate `eps` into `min_branch_length`

### LINKED ISSUE #483: Default epsilon is occasionally too small
In edge cases (e.g. with a single EP iteration) the epsilon used to force positive branch lengths is too small, and gets lost due to floating point error, causing the subsequent `tables.sort()` to error out. In general, a "safe" lower bound for epsilon is going to depend on the mutational clock & time scale: that is, if the mutation rate is very low and/or population size very large then we might run into issues even outside edge cases.

It would be nice to (1) document the `eps` argument somewhere so it's more clear how to work around issues like this (and maybe rename it to be clearer; something like `min_branch_length`); (2) if `eps` falls below the minimum floating point difference for a particular edge length, use a larger value (but still small in relative terms) so that we don't error out.

nspope:
And change default to 1e-8. Closes #483

### DIFF
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -2064,13 +2064,13 @@ class TestPopulationSizeHistory:
         beta = 1.7
         demography = PopulationSizeHistory([1000, 2000, 3000], [500, 2500])
         numer_mn, _ = scipy.integrate.quad(
-            lambda t: demography.to_natural_timescale(np.array([t]))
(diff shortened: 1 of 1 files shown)


nspope:
Tests need some fixes; some (not related to this PR) fail for tskit 1.0 b/c of mutation parents

PR #492, opened 2026-01-26
Add preprint citation to README

nspope:


nspope:
Anything more to say here do you think, @hyanwong ?

PR #495, opened 2026-02-27
New layout

jeromekelleher:
Brings the repo into the same layout as the rest of tskit-dev

Still a WIP

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,4 +10,4 @@ on:
 
 jobs:
   Docs:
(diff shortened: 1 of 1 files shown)


jeromekelleher:
> One thing I don't have my head around: will uv.lock need to be updated via uv lock in a commit whenever modifying dependencies in pyproject.toml, for the CI to work?

Yes, exactly. The idea is that you update the lock file every six months or so, deal with any breakage, and commit. Everything is locked down in between those updates so there's no noise and things stay working.

